Issue #27, opened 2025-05-19
Block reward calculator


sandtreader:
A new module that accepts block bodies and calculates the rewards available on that block, using:

- Total fees from transactions
- Protocol parameters

Output on cardano.block.rewards indicates which SPO produced the block and the total rewards issued.  It should also issue 'pot' changes for the element which is withdrawn from reserves.

sandtreader:
Rethink:  This only need to send data at the end of an epoch.  It should send the total fees available and a list of KES public keys used for each block.  The reward calculator can then use the parameter set to get ρ and τ to calculate the rewards available, and match the KES keys against the SPO set (from SPOState) to total the blocks produced by each SPO.

This means the BlockRewardCalculator should probably be called EpochActivityCounter.

PR #30, opened 2025-05-20
Prc/27 epoch activity counter

sandtreader:
Epoch Activity Counter (#27) - counts total fees and which SPOs (by VRF key at this stage) successfully minted blocks.

Note rollbacks are not yet handled and there is a race between the two subscriptions - fees should be processed first, because the headers trigger the epoch event.  Useful test case on how to resolve generally.


alexwoods (approved):
Looks good to me

Issue #31, opened 2025-05-21
Refactor Messages to have a sub-enum for CardanoMessages with BlockInfo


sandtreader:
To allow generic treatment of Cardano messages (which currently independently include BlockInfo) we want to define a CardanoMessage which includes BlockInfo and then contains an enum of messages which don't.  This will then be in a Cardano variant of the outer Message.

Impact will be a change to the `match message` on every subscriber to add the Cardano() level.

PR #32, opened 2025-05-21
#31 Encapsulate Cardano messages in a sub-enum

sandtreader:
Moves BlockInfo into a common place which can be access for (e.g.) synchronisation


alexwoods (approved):
Looks good to me

PR #42, opened 2025-06-11
Message-Based Testing

yHSJ:
This PR introduces:
- a new `ledger_state` type that handles the loading and decoding of a canonical ledger state (to be defined)
- a new `snapshot_bootstrapper` module that constructs the above `ledger_state` and broadcasts `SnapshotState` messages with each element of the `ledger_state`
- logic that allows the `spo_state` module to load state from a `SnapshotState` message

TODO:
- Figure out how to apply blocks and then fetch the state once blocks have been applied
- Write tests that orchestrate all these components together


common/src/messages.rs
```diff
@@ -171,6 +178,9 @@ pub enum Message {
 
     // Cardano messages with attached BlockInfo
     Cardano((BlockInfo, CardanoMessage)),
+
+    // Initialize state from a snapshot
+    SnapshotState(SnapshotStateMessage),
```
SupernaviX, line 183: I just tested on my machine, and I hit the same issues you were seeing adding `#[cfg(test)]` here. It looks like (if we want conditional compilation) we need to use a custom feature instead.

common/Cargo.toml
```diff
@@ -26,6 +26,8 @@ serde_json = "1.0"
 serde_with = { version = "3.12.0", features = ["hex"] }
 tokio = { version = "1", features = ["full"] }
 tracing = "0.1.40"
+minicbor = { version = "0.26.0", features = ["std", "half", "derive"] }
```
sandtreader, line 29: We already use serde::cbor in Caryatid for message encoding - I'm wary of adding another different crate to do the same, and having conflicts in encoding decorations.  Is there anything we need specifically from minicbor that serde::cbor can't do?

common/src/types.rs
```diff
@@ -149,6 +150,33 @@ pub struct Ratio {
     pub denominator: u64,
 }
 
+// NOTE: Implementations from Pallas, why aren't we just using Pallas types here?
```
sandtreader, line 153: Strategic decision not to make common (and hence every module) dependent on Pallas, for a number of reasons:

1. Keeping dependencies to a minimum
2. Version control
3. Keeping control of our message encoding

This does cause us some pain in reimplementing some things but long term gains, I believe.

common/src/types.rs
```diff
@@ -149,6 +150,33 @@ pub struct Ratio {
...
+impl<'b, C> minicbor::decode::Decode<'b, C> for Ratio {
```
sandtreader, line 154: Actually we already have a RationalNumber (common/src/rational_number.rs) - not sure why the PoolRegistration doesn't use it.  This encoding (if required) should move there I think - I'd like to keep types.rs as free of implementation as possible.

modules/spo_state/src/spo_state.rs
```diff
@@ -43,8 +50,33 @@ impl SPOState {
...
         info!("Creating SPO state publisher on '{spo_state_topic}'");
 
+        let can_bootstrap_from_snapshot = config
+            .get_bool("can-bootstrap-from-snapshot")
+            .unwrap_or(false);
+        info!("Able to use snapshot to bootstrap: {can_bootstrap_from_snapshot}");
+
         let state = Arc::new(Mutex::new(State::new()));
 
+        // Subscribe for snapshot messages
```
sandtreader, line 60: Just don't subscribe at all if not allowed - reduces the number of subscriptions in the message bus

modules/spo_state/src/state.rs
```diff
@@ -42,6 +44,23 @@ impl BlockState {
     }
 }
 
+impl From<SPOState> for BlockState {
```
sandtreader, line 47: Nice :-)

modules/spo_state/src/state.rs
```diff
@@ -109,6 +155,16 @@ impl State {
         }
     }
 
+    /// Returns a reference to the block state at a specified height, if applicable
```
sandtreader, line 158: Worth noting that there will only be (k=2160) blocks in the history - anything older than that is thrown away.  So great if you just want to capture the tip without race conditions, but you can't use for historical stuff.

processes/golden_tests/golden.toml
```diff
@@ -0,0 +1,71 @@
...
+port = 4340
+
+[module.spy]
+# Enable for message spying
+#topic = "cardano.stake.deltas"
+
+[startup]
+topic = "cardano.sequence.start"
+
+[message-bus.external]
```
sandtreader, line 55: Don't need this - generally I wonder if the config should be the minimum necessary to run the tests we've actually implemented so far...

ledger-state/cddl/common.cddl
```diff
@@ -0,0 +1,31 @@
+; This file contains some some basic, shared types that are used throughout the ledger state specification.
+
+hash_28 = bytes .size 32
```
SupernaviX, line 3: Forgive my ignorance, but is this size a typo?

yHSJ:
I've marked this PR as "Ready for Review" now, even though it's clearly not done. I left some notes for whoever picks up this work once we get time to focus back on Acropolis in commit https://github.com/input-output-hk/acropolis/pull/42/commits/0e0a742b7603ffa4930ba1ee47eaeb982823c2ae

Merging this will make it easier to pick back up in the future, less "loose ends" to tie up in N months once we forget everything.

@sandtreader If you think there is any other documentation or notes I should leave, please let me know. Otherwise, feel free to merge this whenever you think the time is right.

PR #47, opened 2025-06-17
Genesis now parsed by parameters-state module

shd:
1. Genesis parsing moved from bootstrapper to parameters-state
2. Added byron and shelley genesis parameter sets


common/src/rational_number.rs
```diff
@@ -61,6 +62,24 @@ impl From<u64> for RationalNumber {
     }
 }
 
+impl TryFrom<f32> for RationalNumber {
```
sandtreader, line 65: I'm a bit worried about why this is necessary - are rational numbers ever encoded as float?  If so there could be loss of precision and will regenerate as a massive fraction, I guess (1/3 => 0.33333..333 => 33333..333/100000..000)

modules/genesis_bootstrapper/src/genesis_bootstrapper.rs
```diff
@@ -23,8 +18,6 @@ const DEFAULT_COMPLETION_TOPIC: &str = "cardano.sequence.bootstrapped";
 
 // Include genesis data (downloaded by build.rs)
 const MAINNET_BYRON_GENESIS: &[u8] = include_bytes!("../downloads/mainnet-byron-genesis.json");
-const MAINNET_SHELLEY_GENESIS: &[u8] = include_bytes!("../downloads/mainnet-shelley-genesis.json");
```
sandtreader, line 26: These need removing from its build.rs too

modules/parameters_state/src/genesis_params.rs
```diff
@@ -0,0 +1,256 @@
...
+pub fn map_fraction(fraction: &conway::Fraction) -> RationalNumber {
+    RationalNumber {
+        numerator: fraction.numerator,
+        denominator: fraction.denominator,
+    }
+}
+
+fn map_pool_thresholds(thresholds: &conway::PoolVotingThresholds) -> Result<PoolVotingThresholds> {
+    Ok(PoolVotingThresholds {
+        motion_no_confidence: RationalNumber::try_from(thresholds.motion_no_confidence)?,
```
sandtreader, line 50: Ah, here we are - is Pallas flattening the rational number into a float?  That seems like a fault in Pallas if so.

modules/genesis_bootstrapper/src/genesis_bootstrapper.rs
```diff
@@ -224,19 +93,11 @@ impl GenesisBootstrapper {
...
-                        error!("Failure to parse conway genesis block: {e}");
-                        None
-                    }),
-            };
-
             let message_enum = Message::Cardano((
                 block_info,
-                CardanoMessage::GenesisComplete(completion_message),
+                CardanoMessage::GenesisComplete(GenesisCompleteMessage { 
+                    conway_genesis: None
```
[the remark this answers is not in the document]
shd: I do not use "GenesisComplete" anymore, so maybe it could be done

modules/parameters_state/src/genesis_params.rs
shd: Unluckily it just reads it from Genesis

PR #49, opened 2025-06-18
DRep Delegation Distribution

### LINKED ISSUE #41: Generate total stake delegated to DReps
The DRepState currently generates the DRepStakeDistribution message using the DRep's deposit - this was a placeholder, and is not what's needed.  It should generate the total stake delegated to each DRep, which requires tracking both vote delegation certificates (all 4 types!) and also stake address values.

AccountsState already calculates stake address values so DRepState could ask it for the total value for a set of stake addresses delegated to a particular DRep - this is likely to be quite a bit of messenging for a few million delegated addresses.

Alternatively, DRepState could duplicate the stake address counting from AccountsState and hold it itself.  With 5 million stake addresses if it only holds the utxo_value then it's not too bad in memory.   We will need a nice data structure which can represent the DRep -> StakeAddress and StakeAddress -> stake-value maps efficiently.  Consider 

StakeAddress -> index
[utxo_value] by index
DRep -> [index]

Where index is incremented for each new StakeAddress seen.  This saves storing StakeAddress twice and should give faster lookup.

SupernaviX:
Compute drep delegation distribution in the account state module. Every epoch, we generate the DRDD and yeet it to whoever's listening. We also have a REST endpoint which exposes it.

Closes #41

### DIFF
--- a/common/src/messages.rs
+++ b/common/src/messages.rs
@@ -115,8 +115,17 @@ pub struct DRepStateMessage {
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 pub struct DRepStakeDistributionMessage {
-    /// DRep stake distribution by ID, if None then no changes
-    pub data: Option<Vec<(DRepCredential, Lovelace)>>,
+    /// Epoch which has ended
+    pub epoch: u64,
+
+    /// DRep stake assigned to the special "abstain" DRep.
+    pub abstain: Lovelace,
+
+    /// DRep stake assigned to the special "no confidence" DRep
+    pub no_confidence: Lovelace,
+
+    /// DRep stake distribution by ID
+    pub dreps: Vec<(DRepCredential, Lovelace)>,
 }
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
--- a/common/src/types.rs
+++ b/common/src/types.rs
@@ -188,6 +188,13 @@ impl Credential {
         }
     }
 
+    pub fn to_json_string(&self) -> String {
+        match self {
+            Self::ScriptHash(hash) => format!("scriptHash-{}", hex::encode(hash)),
+            Self::AddrKeyHash(hash) => format!("keyHash-{}", hex::encode(hash)),
+        }
+    }
+
     pub fn get_hash(&self) -> KeyHash {
         match self {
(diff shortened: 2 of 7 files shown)


modules/accounts_state/src/state.rs
```diff
@@ -93,6 +108,60 @@ impl State {
...
+    /// Derive the DRep Delegation Distribution (SPDD) - the total amount
+    /// delegated to each DRep, including the special "abstain" and "no confidence" dreps.
+    pub fn generate_drdd(&self) -> DRepDelegationDistribution {
+        let abstain = AtomicU64::new(0);
+        let no_confidence = AtomicU64::new(0);
+        let dreps = self
+            .dreps
+            .iter()
+            .map(|(cred, deposit)| (cred.clone(), AtomicU64::new(*deposit)))
+            .collect::<BTreeMap<_, _>>();
```
SupernaviX, line 120: This is using an immutable `BTreeMap<_, AtomicU64>` instead of a `DashMap<_, u64>`. That's because dashmap uses locking under the hood; it partitions entries into some number of "shards", and each shard is guarded by a `RwLock`.

common/src/messages.rs
```diff
@@ -115,8 +115,17 @@ pub struct DRepStateMessage {
...
+    pub epoch: u64,
+
+    /// DRep stake assigned to the special "abstain" DRep.
+    pub abstain: Lovelace,
+
+    /// DRep stake assigned to the special "no confidence" DRep
+    pub no_confidence: Lovelace,
+
+    /// DRep stake distribution by ID
+    pub dreps: Vec<(DRepCredential, Lovelace)>,
```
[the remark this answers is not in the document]
sandtreader: That's right, it should be the full state each time - the synchronisation across multiple topics relies on this.

common/src/types.rs
```diff
@@ -188,6 +188,13 @@ impl Credential {
         }
     }
 
+    pub fn to_json_string(&self) -> String {
+        match self {
+            Self::ScriptHash(hash) => format!("scriptHash-{}", hex::encode(hash)),
+            Self::AddrKeyHash(hash) => format!("keyHash-{}", hex::encode(hash)),
```
sandtreader, line 194: I wonder if there's some base58-style convention for these?

PR #50, opened 2025-06-25
Shd/governance state

shd:
Rewards and refunds


common/src/messages.rs
```diff
@@ -132,8 +132,10 @@ pub struct ProtocolParamsMessage {
 }
 
 #[derive(Debug, Default, Clone, serde::Serialize, serde::Deserialize)]
-pub struct EnactStateMessage {
-    pub enactments: Vec<EnactStateElem>,
+pub struct GovernanceOutcomesMessage {
```
sandtreader, line 135: Looks good - I wonder if we should add stuff about the GA itself - its ID, metadta, success/failure reason, voting outcomes,...? for other apps that want governance data.

shd: That's a lot of info! Maybe I can add GovActionId, boolean flag (accepted/rejected) and voting stats (total number of cast votes)? If there will be additional request, we can always add it. 

A second possible idea is to add Proposal. And I'm afraid to add the voting info (total transcripts etc).

common/src/messages.rs
```diff
@@ -131,9 +131,17 @@ pub struct ProtocolParamsMessage {
...
+/// action. If the action is not accepted or has no associated state change (like
+/// Information), then it is included into `refunds` field. Otherwise info is 
+/// specified in `enact_state`/`withdrawals` field and not repeated in `refunds`.
 #[derive(Debug, Default, Clone, serde::Serialize, serde::Deserialize)]
-pub struct EnactStateMessage {
-    pub enactments: Vec<EnactStateElem>,
+pub struct GovernanceOutcomesMessage {
+    pub enact_state: Vec<(VotingOutcome, EnactStateElem)>,
+    pub withdrawals: Vec<(VotingOutcome, TreasuryWithdrawalsAction)>,
+    pub refunds: Vec<VotingOutcome>
```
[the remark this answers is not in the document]
shd: VotingOutcome is a useful structure, I'd like to keep it as a separate one, if there is no other reason.

But I can reorder the structs in this way, ok
```GovernanceOutcome {
     outcome: VotingOutcome,
     enact_state: ...
     withdrawals: ...
}```

It is not allowed to have them simultaneously (state/withdrawal), these are two distinct actions. But it's not important, it does not break anything

PR #66, opened 2025-07-08
feat: Update `handle_rest_with_parameter` to extract REST params dynamically

whankinsiv:
This PR:
* Adds `extract_params_from_topic_and_path` to dynamically extract parameters by matching `*` positions in the topic, replacing hardcoded parameter positions.
* Updates existing REST endpoints to work with the new `&[&str]` parameter type of `handle_rest_with_parameter`.

### DIFF
--- a/common/src/rest_helper.rs
+++ b/common/src/rest_helper.rs
@@ -49,23 +49,30 @@ pub fn handle_rest_with_parameter<F, Fut>(
     handler: F,
 ) -> JoinHandle<()>
 where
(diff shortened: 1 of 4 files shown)


sandtreader:
Neat idea!  I think it should be '#' though, since that matches a single 'word'.  '*' matches everything following, so maybe that should capture all the rest into a single param?

e.g. for topic `rest.get.foo.bar.splat`

```
foo.#.# => { 'bar', 'splat' }
foo.* => { 'bar.splat' }

```
This will be backwards incompatible with topics requesting '.*' but they probably mean '.#' anyway.

PR #68, opened 2025-07-08
Refactor existing `governance-state` REST endpoint and add proposal info and votes endpoints

whankinsiv:
This PR:
* Refactors the existing `/governance-state/list` endpoint to use `handle_rest` and return only `GovActionId` in Bech32 format, instead of `GovActionId` objects and all votes stored in state. 
* Adds `/governance-state/proposal/<Bech32_GovActionId>` endpoint to return deposit, rewards_account, gov_action_id, gov_action, and anchor for a specific proposal. 
* Adds `/governance-state/proposal/<Bech32_GovActionId>/votes` endpoint to return all votes (`DataHash` and `VotingProcedures`) keyed by voter.

### DIFF
--- a/modules/governance_state/src/governance_state.rs
+++ b/modules/governance_state/src/governance_state.rs
@@ -6,12 +6,12 @@ use acropolis_common::{
         CardanoMessage, DRepStakeDistributionMessage, GovernanceProceduresMessage, Message,
         ProtocolParamsMessage, RESTResponse,
     },
-    BlockInfo,
+    rest_helper::{handle_rest, handle_rest_with_parameter},
+    BlockInfo, GovActionId,
 };
 use anyhow::{anyhow, Result};
 use caryatid_sdk::{message_bus::Subscription, module, Context, Module};
 use config::Config;
-use hex::ToHex;
 use std::sync::Arc;
 use tokio::sync::Mutex;
 use tracing::{error, info};
@@ -37,7 +37,7 @@ pub struct GovernanceState;
 
 pub struct GovernanceStateConfig {
     subscribe_topic: String,
-    handle_topic: String,
+    _handle_topic: String,
     drep_distribution_topic: String,
     protocol_parameters_topic: String,
     enact_state_topic: String,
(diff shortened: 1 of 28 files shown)


modules/governance_state/src/governance_state.rs
```diff
@@ -58,7 +59,7 @@ impl GovernanceStateConfig {
     pub fn new(config: &Arc<Config>) -> Arc<Self> {
         Arc::new(Self {
             subscribe_topic: Self::conf(config, DEFAULT_SUBSCRIBE_TOPIC),
-            handle_topic: Self::conf(config, DEFAULT_HANDLE_TOPIC),
+            _handle_topic: Self::conf(config, DEFAULT_HANDLE_TOPIC),
```
sandtreader, line 62: I think we still want the topics to be configurable, although there may need to be more than one config rather than trying to construct them.

modules/governance_state/src/governance_state.rs
```diff
@@ -207,13 +144,188 @@ impl GovernanceState {
...
+                }
+            },
+        );
+
+        // Handle REST requests for /governance-state/proposal/<Bech32_GovActionId>
+        let state_handle_proposal_info = state.clone();
+        handle_rest_with_parameter(
+            context.clone(),
+            "rest.get.governance-state.proposal.*",
+            move |param| {
```
sandtreader, line 193: I wonder if now the closures are getting quite long, we could split them out into functions in a separate file?

modules/governance_state/src/governance_state.rs
```diff
@@ -207,13 +144,188 @@ impl GovernanceState {
...
+                        ),
+                            ));
+                        }
+                    };
+
+                    let locked = state.lock().await;
+                    match locked.get_proposal(&proposal_id) {
+                        Some(proposal) => {
+                            #[derive(serde::Serialize)]
+                            struct ProposalProcedureRest {
```
sandtreader, line 213: If in a separate file these structs could be at the top level too

modules/governance_state/src/state.rs
```diff
@@ -400,28 +400,27 @@ impl State {
     }
 
     /// Get list of actual voting proposals
-    pub fn list_proposals(&self) -> Result<Vec<(GovActionId, ProposalProcedure)>> {
-        let mut result = Vec::new();
-        for (action, (_epoch, prop)) in self.proposals.iter() {
-            result.push((action.clone(), prop.clone()))
-        }
-        Ok(result)
+    pub fn list_proposals(&self) -> Vec<GovActionId> {
```
sandtreader, line 403: Nice :+1:

sandtreader (changes requested):
Generally great - just would like to retain configurable topics and maybe split the REST implementations into another file

modules/governance_state/src/governance_state.rs
whankinsiv: Moved to newly defined `rest.rs` inside `/modules/governance_state/src/`

modules/governance_state/src/governance_state.rs
whankinsiv: Defined a topic for each endpoint to allow for configurability.

sandtreader:
Yes really neat now, thanks!

PR #70, opened 2025-07-09
Shd/replayer update

shd:
Added recorder/playback modules directly into replayer sub-project.
Allows to collect governance channels, and then play it back.
This is a specific solution for governance: maybe some general solution could be extracted from this code.

### DIFF
--- a/common/src/messages.rs
+++ b/common/src/messages.rs
@@ -108,7 +108,7 @@ pub struct EpochActivityMessage {
     pub vrf_vkey_hashes: Vec<(KeyHash, usize)>,
 }
 
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Default, Clone, serde::Serialize, serde::Deserialize)]
 pub struct GovernanceProceduresMessage {
     /// Proposals
     pub proposal_procedures: Vec<ProposalProcedure>,
@@ -126,7 +126,7 @@ pub struct DRepStateMessage {
     pub dreps: Vec<(DRepCredential, Lovelace)>,
 }
 
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Default, Clone, serde::Serialize, serde::Deserialize)]
 pub struct DRepStakeDistributionMessage {
     /// Epoch which has ended
     pub epoch: u64,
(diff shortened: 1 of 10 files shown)


common/src/messages.rs
```diff
@@ -108,7 +108,7 @@ pub struct EpochActivityMessage {
     pub vrf_vkey_hashes: Vec<(KeyHash, usize)>,
 }
 
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Default, Clone, serde::Serialize, serde::Deserialize)]
```
sandtreader, line 111: I'm a bit reluctant to add Default back to all the messages - I removed it a while back because there weren't actually sensible defaults for every message...  Can this be avoided?  I see it's used to construct empty messages, but would Message::None() do the same?

common/src/messages.rs
```diff
@@ -223,9 +223,9 @@ pub enum SnapshotStateMessage {
 }
 
 // === Global message enum ===
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Default, Clone, serde::Serialize, serde::Deserialize)]
 pub enum Message {
-    None(()), // Just so we have a simple default
+    #[default] None, // Just so we have a simple default
```
sandtreader, line 228: I learned something today :-)

processes/replayer/src/playback.rs
```diff
@@ -0,0 +1,225 @@
...
+        }
+    }
+}
+
+impl Playback {
+    async fn init(&self, context: Arc<Context<Message>>, config: Arc<Config>) -> Result<()> {
+        let cfg = ReplayerConfig::new(&config);
+        let mut playback_runner = PlaybackRunner::new(context, cfg);
+
+        tokio::spawn(async move {
```
sandtreader, line 59: FYI there's a Context.run() which does this for you, and also ensures everything is ready before it starts.

common/src/messages.rs
shd: Maybe we can add something like "CardanoMessage::Empty" to "CardanoMessage". I'm not sure that this is the excellent solution, but "Empty" message can be processed in all topics and be considered as allowed.

shd: (although the meaning of this Default is different: to have a simple constructor for the message)

shd:
Made two small fixes. Probably, we can merge it now

PR #72, opened 2025-07-14
DRep state REST refactor

whankinsiv:
This PR:
* Adds `from_drep_bech32` and `to_drep_bech32` helper methods to `Credential` for converting between DRep Bech32 strings and `Credential` objects.
* Renames the `/drep-state/list` endpoint to `/dreps/list` for cleaner REST API design. 
* Consolidates `/drep-state/address=` and `/drep-state/script=` into a single `/dreps/info/{drep_id}` endpoint using Bech32 DRep identifiers instead of hex credentials.
* Refactors both endpoints to use the `rest_helper` handlers to reduce duplication.
* Moves both REST handler implementations to a new dedicated `rest.rs` file.

### DIFF
--- a/common/src/serialization.rs
+++ b/common/src/serialization.rs
@@ -1,3 +1,4 @@
+use anyhow::anyhow;
 use bech32::{Bech32, Hrp};
 use serde::{ser::SerializeMap, Serializer};
 use serde_with::{ser::SerializeAsWrap, SerializeAs};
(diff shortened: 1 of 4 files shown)


sandtreader:
Interesting debate about REST structure - should be it be:

dreps/list
dreps/info/{id}

or just

dreps
dreps/{id}

The latter feels more 'RESTy' :-)

PR #78, opened 2025-07-15
Improve REST refactor & OpenAPI specification

whankinsiv:
This PR:
* Fixes endpoint return types to use Bech32 encoding wherever applicable.
* Modifies REST endpoints to use `/foo` and `/foo/{id}` paths instead of `/foo/list` and `/foo/info/{id}` format across all modules.
* Moves `accounts`, `utxo`, and `spo` REST handlers into dedicated `rest.rs` files, matching the structure of the `drep` and `governance` modules.
* Adds a new `/pools` endpoint to return an object keyed by Bech32 pool IDs, including margin, fixed cost, and pledge information.
* Expands the OpenAPI specification to cover all supported relay types for `/pools/{pool_id}` endpoint.

### DIFF
--- a/API/openapi.yaml
+++ b/API/openapi.yaml
@@ -21,7 +21,7 @@ tags:
     description: Endpoints related to UTxO state.
 
 paths:
-  /accounts/info/{stake_address}:
+  /accounts/{stake_address}:
     get:
       tags:
         - Accounts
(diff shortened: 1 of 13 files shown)


sandtreader:
Great stuff!

Issue #87, opened 2025-07-16
Historical REST interface for ParameterState


sandtreader:
Add an optional history of ParameterState REST interface

/parameters/epoch/<n>

Issue #86, opened 2025-07-16
Fix REST interface in ParameterState


sandtreader:
Reinstate simple REST interface to ParameterState providing the current state

/parameters/

PR #90, opened 2025-07-18
feat: Add protocol parameter REST endpoints with optional historical storage (WIP)

whankinsiv:
Opening as a draft to get initial feedback. Updating OpenAPI specification and final testing will be done tomorrow.
This PR:
* Adds `active_epoch` and optional `parameter_history` fields to `parameters_state`.
* Introduces a `store-history` config option in `omnibus.toml` to enable historical protocol parameters storage.
* Implements `/epoch/parameters` and `/epochs/{epoch_number}/parameters` REST endpoints mentioned in #86 & #87.

### DIFF
--- a/modules/parameters_state/src/parameters_state.rs
+++ b/modules/parameters_state/src/parameters_state.rs
@@ -2,7 +2,8 @@
 //! Accepts certificate events and derives the Governance State in memory
 
 use acropolis_common::{
-    messages::{CardanoMessage, Message, ProtocolParamsMessage, RESTResponse},
+    messages::{CardanoMessage, Message, ProtocolParamsMessage},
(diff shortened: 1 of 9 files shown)


whankinsiv:
Merge conflicts are now resolved. Moving to ready for review.

modules/parameters_state/src/parameters_state.rs
```diff
@@ -48,13 +61,21 @@ impl ParametersStateConfig {
         actual
     }
 
+    fn conf_bool(config: &Arc<Config>, keydef: (&str, bool)) -> bool {
```
sandtreader, line 64: Thinking these config helpers should be in common really.  If only we could attach helpers to Config itself!

Issue #95, opened 2025-07-24
Create `/pools/retiring` endpoint


golddydev:
The `spo-state` module stores `pending_deregistrations` as a map of epochs numbers to pool_ids. 

### Reference

https://docs.blockfrost.io/#tag/cardano--pools/GET/pools/retiring

### Acceptance Criteria

- [ ] Add a new REST endpoint `/pools/retiring` that returns a JSON object where each key is an epoch number and the value is  a list of Bech32-encoded pool IDs scheduled to retire in that epoch. (Return type must match with BlockfrostAPI Specification)
- [ ] Add the OpenAPI schema entry for  `/pools/retiring` to `openapi.yaml`

PR #96, opened 2025-07-24
feat: add /pools/retiring endpoint

golddydev:
## Summary

This PR adds `/pools/retiring` endpoint to get all pools which will be retired in upcoming epochs.

## Changes

* Add `/pools/retiring` endpoint to `spo_state` module's `spo` rest handler.
* Add corresponding OpenAPI spec with `PoolRetirement` schema.
* Add unit tests for `get_retiring_pools` function.

## Related Issue

#95

### DIFF
--- a/modules/spo_state/src/state.rs
+++ b/modules/spo_state/src/state.rs
@@ -585,11 +585,12 @@ pub mod tests {
         assert!(state.handle_tx_certs(&block, &msg).is_ok());
         let current = state.current();
         assert!(!current.is_none());
-        let retiring_pools = state.get_retiring_pools();
+        let mut retiring_pools = state.get_retiring_pools();
+        retiring_pools.sort_by_key(|p| p.epoch);
         assert_eq!(2, retiring_pools.len());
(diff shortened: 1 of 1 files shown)


modules/spo_state/src/state.rs
```diff
@@ -114,6 +114,24 @@ impl State {
         self.current().map(|state| state.spos.iter().collect())
     }
 
+    /// Get pools that will be retired in the upcoming epochs
+    pub fn get_retiring_pools(&self) -> Option<Vec<PoolRetirement>> {
```
sandtreader, line 118: Just wondering why this is an Option<> - feels like that's an implementation detail and it could return a (definite) empty Vec if there aren't any?

golddydev: Yeah, that makes sense.  👍🏼

golddydev: [3733f96](https://github.com/input-output-hk/acropolis/pull/96/commits/3733f96a15a9c8e1063b787edaf60c1c8cfc8b5a)

PR #104, opened 2025-07-29
refactor: Mithril fetcher to download intelligently

golddydev:
### Overview
This PR introduces an intelligent snapshot download system that manages Mithril snapshot downloads more efficiently based on age constraints, replacing the simple boolean `download` flag with a more granular `download-max-age` configuration.

#### **Age-Based Download Logic**
- **unset or invalid `u64`**: Never download.
- **`0`**: Always download (equivalent to old `download = true`)
- **`8`**: Download only if there is no snapshot within 8 hours

#### **Metadata Persistence**
- Saves snapshot metadata as JSON for reliable age tracking
- Uses snapshot creation time from metadata (not file timestamp)
- Enables intelligent age comparison between snapshots

#### **Smart Download Decisions**
- Compares existing snapshot age against `download-max-age` configuration
- Only downloads if current snapshot is expired AND newer snapshot is available
- Prevents unnecessary downloads when valid snapshots exist

#### **Dependencies**
- Added `chrono = "0.4.41"` for time handling
- Added `Snapshot` import from `mithril_client`

#### **New Functions**
```rust
fn load_snapshot_metadata(path: &Path) -> Result<Snapshot>
fn save_snapshot_metadata(snapshot: &Snapshot, path: &Path) -> Result<()>
fn should_skip_download(old_snapshot: &Snapshot, latest_snapshot: &Snapshot, config: &Config) -> bool
```

#### **Configuration Migration**
```toml
# Before
download = false

# After
download-max-age = "never"  # Never download or
download-max-age = 0      # Always download  or
download-max-age = 8      # Download if older than 8 hours
```

### 🧪 Testing
- Added comprehensive unit tests for metadata persistence
- Added tests for age-based download logic

## Related Issue
#2

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -136,6 +136,7 @@ dependencies = [
  "anyhow",
  "async-trait",
  "caryatid_sdk",
+ "chrono",
  "config",
  "mithril-client",
  "pallas",
--- a/modules/mithril_snapshot_fetcher/Cargo.toml
+++ b/modules/mithril_snapshot_fetcher/Cargo.toml
@@ -19,6 +19,7 @@ tracing = "0.1.40"
 serde_json = "1.0.138"
 mithril-client = { version = "0.12", features = ["fs"] }
 async-trait = "0.1.86"
+chrono = "0.4.41"
 
 [build-dependencies]
 reqwest = { version = "0.11", features = ["blocking"] }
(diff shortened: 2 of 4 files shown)


sandtreader:
Excellent update and PR docs, no notes!

PR #122, opened 2025-08-21
feat: Expanded drep_state with historical state and rollback support

whankinsiv:
This PR expands the `drep_state` module to support all state data needed to align with all /governance/dreps/ related Blockfrost endpoints. A follow up PR will wire these endpoints to the new `historical_dreps` data. 

Main changes:
* Added`HistoricalDRepState` keyed by `DRepCredential`  storing:
   * extended DRep info (deposit, expired/retired flags, active epoch)
   * updates history  (registrations, deregistrations, updates with `tx_hash` and `cert_index`)
   * metadata anchors
   * delegators credentials
   * votes records
* Uses `StateHistory` to commit state changes by block, enabling rollback support. 
* Added conditional subscribers: 
   * `cardano.governance` (for votes when `store-votes` enabled)
   * `cardano.protocol.parameters` (for drep expiration tracking when `store-info` enabled)
* Extended `GovernanceProceduresMessage` to include `vote_index` in each `VotingProcedure`.
* Extended `DRepRegistration`, `DRepDeregistration`, and `DRepUpdate` types of `TxCertificate` with `tx_hash` and `cert_index`.

This implementation makes storage configurable within the module so users can keep resource costs low when DRep data isn't needed.

### DIFF
--- a/.github/workflows/run-tests-on-push-to-main.yml
+++ b/.github/workflows/run-tests-on-push-to-main.yml
@@ -14,13 +14,17 @@ env:
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
 
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
-      - name: Build
+
+      - name: Run Format
+        run: cargo fmt --all -- --check
+
+      - name: Run Build
         run: cargo build --verbose
+
       - name: Run tests
         run: cargo test --verbose
--- a/common/src/calculations.rs
+++ b/common/src/calculations.rs
@@ -11,7 +11,9 @@ pub fn slot_to_epoch(slot: u64) -> u64 {
 }
 
 pub fn slot_to_epoch_with_shelley_params(
-    slot: u64, shelley_epoch: u64, shelley_epoch_len: u64
+    slot: u64,
+    shelley_epoch: u64,
+    shelley_epoch_len: u64,
 ) -> u64 {
     let shelley_start_slot = shelley_epoch * BYRON_SLOTS_PER_EPOCH;
     if slot < shelley_start_slot {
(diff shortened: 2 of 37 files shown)


common/src/queries/accounts.rs
```diff
@@ -1,5 +1,10 @@
+use std::collections::HashMap;
+
 use crate::{DRepChoice, KeyHash};
 
+pub const DEFAULT_ACCOUNTS_QUERY_TOPIC: (&str, &str) =
+    ("accounts-state-query-topic", "cardano.query.accounts");
+
```
golddydev, line 7: I think this query topic can be used from many different modules (each has different config tables).
Should we make this as constant string? (without any configuration key)

golddydev (approved):
this looks fine to me.

common/src/messages.rs
```diff
@@ -129,7 +129,7 @@ pub struct GovernanceProceduresMessage {
     pub proposal_procedures: Vec<ProposalProcedure>,
 
     /// Voting
-    pub voting_procedures: Vec<(DataHash, VotingProcedures)>,
+    pub voting_procedures: Vec<([u8; 32], VotingProcedures)>,
```
sandtreader, line 132: I wonder if we should just define DataHash as [u8; 32] - or maybe make it more explicitly with SHA256Hash?

common/src/queries/accounts.rs
```diff
@@ -1,5 +1,10 @@
+use std::collections::HashMap;
```
sandtreader, line 1: (fussy) is this needed in accounts.rs itself?

sandtreader (approved):
Great stuff - only comment is whether we should define a hash type.

sandtreader:
Merged but left branch open in case you want to do the hash type

common/src/messages.rs
whankinsiv: I've added a TxHash type alias to [u8;32] in #124

PR #138, opened 2025-09-02
ChainStore module

SupernaviX:
WIP


modules/accounts_state/src/monetary.rs
```diff
@@ -107,12 +107,14 @@ fn calculate_monetary_expansion(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use acropolis_common::protocol_params::{
-        Nonce, NonceVariant, ProtocolVersion, ShelleyProtocolParams,
-    };
     use acropolis_common::rational_number::rational_number_from_f32;
     use acropolis_common::NetworkId;
+    use acropolis_common::{
```
sandtreader: Interesting to see these re-orderings - I assume 'cargo fmt', but why didn't earlier runs on main do this?

modules/accounts_state/src/monetary.rs
```diff
@@ -107,12 +107,14 @@ fn calculate_monetary_expansion(
...
+        protocol_params::{Nonce, NonceVariant, ProtocolVersion, ShelleyProtocolParams},
+        types::GenesisDelegate,
```
sandtreader: We've generally re-exported everything from types - not saying that's right, but should probably be consistent.

modules/chain_store/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "acropolis_module_chain_store"
+version = "0.1.0"
+edition = "2021"
+authors = ["Simon Gellis <simon@sundae.fi>"]
```
sandtreader: + Alex!

modules/chain_store/src/chain_store.rs
```diff
@@ -0,0 +1,557 @@
...
+                let res = Self::handle_blocks_query(&query_store, &state, &query)
+                    .unwrap_or_else(|err| BlocksStateQueryResponse::Error(err.to_string()));
+                Arc::new(Message::StateQueryResponse(StateQueryResponse::Blocks(res)))
+            }
+        });
+
+        let mut new_blocks_subscription = context.subscribe(&new_blocks_topic).await?;
+        let mut params_subscription = context.subscribe(&params_topic).await?;
+        context.run(async move {
+            let mut params_message = params_subscription.read();
```
sandtreader: If this is to clear the genesis one, needs a note because non-obvious

modules/chain_store/src/chain_store.rs
```diff
@@ -0,0 +1,557 @@
...
+        store: &Arc<dyn Store>,
+        state: &State,
+        final_block_is_latest: bool,
+    ) -> Result<Vec<BlockInfo>> {
+        if blocks.is_empty() {
+            return Ok(vec![]);
+        }
+        let mut decoded_blocks = vec![];
+        for block in &blocks {
+            decoded_blocks.push(pallas_traverse::MultiEraBlock::decode(&block.bytes)?);
```
sandtreader: Oh I see you're doing the single decode here - maybe above not an issue, then.

modules/chain_store/src/stores/fjall.rs
```diff
@@ -0,0 +1,377 @@
...
+            return Ok(None);
+        };
+        self.get_by_hash(&hash)
+    }
+
+    fn get_by_number_range(&self, min_number: u64, max_number: u64) -> Result<Vec<Block>> {
+        let min_number_bytes = min_number.to_be_bytes();
+        let max_number_bytes = max_number.to_be_bytes();
+        let mut hashes = vec![];
+        for res in self.block_hashes_by_number.range(min_number_bytes..=max_number_bytes) {
```
sandtreader: Interested to know why you store the hashes instead of just using them directly.

modules/chain_store/src/stores/fjall.rs
```diff
@@ -0,0 +1,377 @@
...
+enum StoredTransaction {
+    #[n(0)]
+    BlockReference {
+        #[n(0)]
+        block_hash: Vec<u8>,
+        #[n(1)]
+        index: usize,
+    },
+    #[n(1)]
+    Inline {
```
sandtreader: Do we ever use this Inline option now?

modules/chain_store/src/stores/fjall.rs
```diff
@@ -0,0 +1,377 @@
...
+        let bytes = test_block_bytes();
+        let info = test_block_info(&bytes);
+        let block = build_block(&info, &bytes);
+
+        state.store.insert_block(&info, &bytes).unwrap();
+
+        let new_block = state.store.get_block_by_number(info.number).unwrap();
+        assert_eq!(block, new_block.unwrap());
+    }
+
```
sandtreader: Test needed for get_blocks_by_number_range

sandtreader (commented):
Great work - comments are primarily about code location

modules/accounts_state/src/monetary.rs
alexwoods: I think it's just because there was additional mod added which moved protocol_params inside the {}

modules/chain_store/src/chain_store.rs
```diff
@@ -0,0 +1,557 @@
...
+
+    fn get_block_by_key(store: &Arc<dyn Store>, block_key: &BlockKey) -> Result<Option<Block>> {
+        match block_key {
+            BlockKey::Hash(hash) => store.get_block_by_hash(hash.as_ref()),
+            BlockKey::Number(number) => store.get_block_by_number(*number),
+        }
+    }
+
+    fn get_block_number(block: &Block) -> Result<u64> {
+        Ok(pallas_traverse::MultiEraBlock::decode(&block.bytes)?.number())
```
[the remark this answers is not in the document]
alexwoods: This is actually used where only the block number is needed to be looked up from the block for /next and /previous calls, in order to save processing the rest of the block. However, it did make me realise that it was happening even when the request used a number rather than a hash, so I've sorted that out!

modules/rest_blockfrost/src/handlers/blocks.rs
```diff
@@ -0,0 +1,778 @@
...
+
+    let block_slot_msg = Arc::new(Message::StateQuery(StateQuery::Blocks(
+        BlocksStateQuery::GetBlockBySlot { slot },
+    )));
+    let block_info = query_state(
+        &context,
+        &handlers_config.blocks_query_topic,
+        block_slot_msg,
+        |message| match message {
+            Message::StateQueryResponse(StateQueryResponse::Blocks(
```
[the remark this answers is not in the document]
alexwoods: I've added a function to common queries utils called rest_query_state. It should be usable for all REST queries I think

PR #150, opened 2025-09-05
Add `timestamp` and `epoch_slot` to `BlockInfo`

### LINKED ISSUE #147: Attach timestamp and epoch_slot to BlockInfo
The timestamp and epoch slot (along with the epoch itself) can be calculated from a slot as long as you have access to certain genesis values. We don't have access to those genesis values in every module, but we should in the modules which introduce blocks to the rest of the system.

Attach these two fields to `BlockInfo` so that we can easily use them elsewhere.

SupernaviX:
Add two useful fields to the common `BlockInfo` struct attached to all Cardano messages: `epoch_slot` and `timestamp`. These fields can both be computed from the `slot`, but only with access to some values populated at genesis. (This is also true for the existing `epoch` field).

This PR also ensures that we compute these values correctly: we create a `GenesisValues` struct in `genesis-bootstrapper`, and read it in modules which need it (`mithril-snapshot-fetcher` and `upstream-chain-fetcher`). This'll make it easier to run in other environments down the road.

@shd if you want to keep running `upstream-chain-fetcher` independently, you need to set `"byron-timestamp"` in your local config. The fetcher still reads genesis values from local config, but it'll wait for `genesis-bootstrapper` to run to use values from there if any of those values are missing from local config.

Fixes #147

### DIFF
--- a/modules/genesis_bootstrapper/build.rs
+++ b/modules/genesis_bootstrapper/build.rs
@@ -1,36 +1,60 @@
 // Build-time script to download generics
-use reqwest::blocking::get;
 use std::fs;
 use std::io::Write;
 use std::path::Path;
 
+use anyhow::{Context, Result};
+
 const OUTPUT_DIR: &str = "downloads";
 
 /// Download a URL to a file in OUTPUT_DIR
-fn download(url: &str, filename: &str) {
-    let response = get(url).expect("Failed to fetch {url}");
-    let data = response.text().expect("Failed to read response");
+async fn download(client: &reqwest::Client, url: &str, filename: &str) -> Result<()> {
+    let request = client.get(url).build().with_context(|| format!("Failed to request {url}"))?;
+    let response =
+        client.execute(request).await.with_context(|| format!("Failed to fetch {url}"))?;
+    let data = response.text().await.context("Failed to read response")?;
 
     let output_path = Path::new(OUTPUT_DIR);
     if !output_path.exists() {
-        fs::create_dir_all(output_path).expect("Failed to create {OUTPUT_DIR} directory");
+        fs::create_dir_all(output_path)
+            .with_context(|| format!("Failed to create {OUTPUT_DIR} directory"))?;
     }
 
     let file_path = output_path.join(filename);
-    let mut file = fs::File::create(&file_path).expect("Failed to create file {file_path}");
-    file.write_all(data.as_bytes()).expect("Failed to write file {file_path}");
+    let mut file = fs::File::create(&file_path)
+        .with_context(|| format!("Failed to create file {}", file_path.display()))?;
+    file.write_all(data.as_bytes())
+        .with_context(|| format!("Failed to write file {}", file_path.display()))?;
+    Ok(())
 }
 
-fn main() {
+#[tokio::main]
+async fn main() -> Result<()> {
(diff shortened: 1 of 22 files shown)


modules/governance_state/src/alonzo_babbage_voting.rs
```diff
@@ -1,30 +1,38 @@
...
+    shelley_slots_per_epoch: u64,
 }
 
 impl AlonzoBabbageVoting {
     pub fn new() -> Self {
-        Self::default()
+        Self {
+            proposals: HashMap::new(),
+            shelley_slots_per_epoch: MAINNET_SHELLEY_SLOTS_PER_EPOCH,
+        }
```
SupernaviX, line 24: This module was hard-coded to use mainnet slots per epoch length, so it wouldn't have worked against sanchonet. I updated it so it defaults to the mainnet value, but reads the real epoch length from protocol parameters when it sees those.

It'd be cleaner to get rid of the default, but I feel like this PR is big enough already. I only updated it in the first place so I could get hard-coded values out of `common`.

modules/genesis_bootstrapper/build.rs
```diff
@@ -1,36 +1,60 @@
...
+        ),
+        download(
+            &client,
+            "https://raw.githubusercontent.com/Hornan7/SanchoNet-Tutorials/refs/heads/main/genesis/byron-genesis.json",
+            "sanchonet-byron-genesis.json",
+        ),
+        download(
+            &client,
+            "https://raw.githubusercontent.com/Hornan7/SanchoNet-Tutorials/refs/heads/main/genesis/shelley-genesis.json",
+            "sanchonet-shelley-genesis.json",
```
shd, line 55: Shelley genesis has bugs, and cannot be properly parsed by Pallas, see parameters-state module.

modules/genesis_bootstrapper/src/genesis_bootstrapper.rs
```diff
@@ -23,7 +24,12 @@ const DEFAULT_NETWORK_NAME: &str = "mainnet";
 
 // Include genesis data (downloaded by build.rs)
 const MAINNET_BYRON_GENESIS: &[u8] = include_bytes!("../downloads/mainnet-byron-genesis.json");
+const MAINNET_SHELLEY_GENESIS: &[u8] = include_bytes!("../downloads/mainnet-shelley-genesis.json");
+const MAINNET_SHELLEY_START_EPOCH: u64 = 208;
```
shd, line 28: Oh, it's still hardcoded... Why not compute it?

modules/genesis_bootstrapper/src/genesis_bootstrapper.rs
```diff
@@ -145,10 +164,16 @@ impl GenesisBootstrapper {
                     .await
                     .unwrap_or_else(|e| error!("Failed to publish: {e}"));
 
+                let values = GenesisValues {
+                    byron_timestamp: byron_genesis.start_time,
+                    shelley_epoch: shelley_start_epoch,
+                    shelley_epoch_len: shelley_genesis.epoch_length.unwrap() as u64,
```
shd, line 170: I'd propose to use parameter-state module output here: it should provide all these data. If it does not, then it's much easier to add additional fields there.

shd (changes requested):

PR #158, opened 2025-09-08
Fixed upstream fetcher wrong epoch bug

shd:
1. Updated upstream-fetcher (now it properly parses the header and properly specifies blocks' era)
2. Updated Alozno voting (now it uses 'update_quorum' field instead of hardcoded value 5).
3. Added Alozno voting test for SanchoNet (it mostly sets only protocol version).
4. Updated documents.


shd:
Appended this PR with fix to Alonzo voting (SanchoNet voting)

PR #171, opened 2025-09-11
feat: CIP25 / 68 metadata storage in assets_state

whankinsiv:
This PR expands `tx_unpacker` and `assets_state` to unpack, publish, process, and store both CIP-25 auxiliary metadata (label 721) and CIP-68 inline datum metadata. With this expansion, all data needed for the `/assets/{asset}` Blockfrost handler is available, and the REST handler is implemented. 

### Main changes:
* Added `datum` to `TxOutput` and `map_datum` helper to convert from Plutus datum to a new enum:
   ```
  pub enum Datum {
      Hash(Vec<u8>),
      Inline(Vec<u8>),
  }
   ```
* Added unpacking of `tx.metadata()` for label 721 (CIP-25) and propagating these blobs via `AssetDeltasMessage`.
* Added `handle_cip25_metadata` in `assets_state` to process CIP-25 auxiliary metadata and store each blob by asset. 
* Extended `assets_state` with an optional subscription to `UTXODeltas` and a `handle_cip68_metadata` handler to process outputs with inline datums and reference tokens (Per CIP68). 
* Implemented `/assets/{asset}` REST handler aligned with Blockfrost's single-asset response. 
* Fixed a bug in `drep_state` which errored when optional history is disabled.


modules/assets_state/src/state.rs
```diff
@@ -315,6 +325,174 @@ impl State {
...
+            supply: self.supply.clone(),
+            history: self.history.clone(),
+            info: new_info,
+            addresses: self.addresses.clone(),
+            transactions: self.transactions.clone(),
+            policy_index: self.policy_index.clone(),
+        })
+    }
+
+    fn resolve_cip68_metadata(
```
sandtreader, line 466: Feels like this very detailed byte-decoding could use some unit tests

modules/rest_blockfrost/src/handlers/assets.rs
```diff
@@ -223,6 +308,142 @@ fn split_policy_and_asset(hex_str: &str) -> Result<(PolicyId, AssetName), RESTRe
     Ok((policy_id, asset_name))
 }
 
+pub async fn fetch_asset_metadata(asset: &str) -> Option<AssetMetadata> {
+    let url = format!(
+        "https://raw.githubusercontent.com/cardano-foundation/cardano-token-registry/master/mappings/{}.json",
```
sandtreader, line 313: This should be configurable (with this as a default).  Is there just one token registry for all networks (mainnet, preprod, preview etc.)?

modules/rest_blockfrost/src/handlers/assets.rs
```diff
@@ -223,6 +308,142 @@ fn split_policy_and_asset(hex_str: &str) -> Result<(PolicyId, AssetName), RESTRe
...
+                }
+            }
+            (Some(Value::Object(obj)), None, None)
+        }
+
+        _ => (None, None, None),
+    }
+}
+
+// TODO: Order fields deterministically to align with blockfrost
```
sandtreader, line 418: TODO in this PR or does it need another issue?

modules/rest_blockfrost/src/handlers/assets.rs
whankinsiv: From my research it appears that there is only one off-chain token registry and its for mainnet.

modules/utxo_state/src/state.rs
```diff
@@ -472,6 +472,7 @@ mod tests {
                     ],
                 )],
             ),
+            datum: None,
```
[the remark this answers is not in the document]
whankinsiv: Created ticket #188

modules/assets_state/src/state.rs
whankinsiv: Test coverage for both metadata handlers and the resolver added in e59d4a3.

processes/omnibus/omnibus.toml
```diff
@@ -8,7 +8,7 @@ genesis-key = "5b3139312c36362c3134302c3138352c3133382c31312c3233372c3230372c323
 # Download max age in hours. E.g. 8 means 8 hours (if there isn't any snapshot within this time range download from Mithril)
 download-max-age = "never"
 # Pause constraint E.g. "epoch:100", "block:1200"
-pause = "none"
+pause = "epoch:470"
```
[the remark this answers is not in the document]
whankinsiv: Set back to `none` in 5ecc15b.

modules/rest_blockfrost/src/handlers/assets.rs
whankinsiv: Updated in 7d5ef1d to use a configurable URL.

Issue #188, opened 2025-09-16
Store TxOutput datum in utxo_state


whankinsiv:
Datums are already unpacked as of #171, but `utxo_state::observe_output()` doesn’t currently persist them. Extend `UTXOValue` to include the datum (supporting both inline and hash forms), and ensure `observe_output()` stores it when outputs are inserted.

PR #237, opened 2025-10-07
Add Datum field to UTXOValue, copy datum into instance of utxovalue w…

lowhung:
This PR extends the UTXO state storage to persist datums, completing the implementation of issue #188. Datums are now stored in both their inline and hash forms via this change.

### Main changes:
* Extend `UTXOValue` to include an optional `datum` field supporting both inline and hash variants via the existing `Datum` enum:
* Updated observe_output() in `utxo_state` to copy the datum from TxOutput when inserting UTXOs into both volatile and immutable storage.
* Added datum assertions to the `observe_output_adds_to_immutable_utxos` test to verify datum persistence through the full storage lifecycle.

Did a small benchmark of the memory footprint for the sync up to epoch 400. This compared the footprint between inclusion of datums and not. For a couple runs I saw 100MB increase (with datums) vs without. This seems reasonable to us.

 Screenshot 2025-10-08 at 2 50 35 PM


whankinsiv (approved):
Looks good. It would be nice to have a benchmark on how storing the datums impacts memory.

PR #241, opened 2025-10-09
feat: address_state module

whankinsiv:
This PR enables optional indexing of address-related data required for conformance with the Blockfrost API specification. 

### Storage lifecycle:
* The module subscribes to address deltas messages emitted by the UTxO state and applies them to the in-memory `volatile` layer which tracks live address updates within the rollback window.
* When the chain advances far enough that an entire previous epoch of deltas can no longer be rolled back, those finalized blocks are pruned from `volatile` and transferred into a pending queue within the `ImmutableAddressStore`.
* A background persistence worker consumes the epoch's pending deltas and atomically writes the updated immutable state to disk. Epochs are processed sequentially, with only one allowed in the queue before blocking the main run loop to prevent unbounded memory growth.

### Retrieving live state:
* The immutable address data is first fetched from Fjall, and any pending immutable entries waiting to be persisted are merged into the result.
* The getter then iterates through each block in the in-memory `volatile` layer, applying any deltas relevant to the queried address on top of the immutable snapshot to produce an up-to-date view of the address state. 

### Data indexed:
```
pub struct AddressEntry {
    // The live utxo set for an address
    pub utxos: Option<Vec<UtxoDelta>>,      
        
    // All transactions involving the address
    pub transactions: Option<Vec<TxIdentifier>>, 

    // The total sent/received amounts for the address (lovelace + native assets)
    pub totals: Option<Vec<ValueDelta>>,            
}
``` 

### REST handlers:
* This PR introduces the REST handler for the `/addresses/{address}` endpoint. 
* A follow-up PR will add the `/addresses/{address}/totals`, `/addresses/{address}/utxos`, and `/addresses/{address}/transactions` endpoints, all of which rely on the data indexed by this module. 

### Next steps:
* Beyond implementing the remaining address-related REST handlers, refactoring the `assets_state` module to adopt the same storage lifecycle will significantly reduce the RAM footprint for address and transaction indexing.


whankinsiv:
All feedback has been addressed. I considered @golddydev's persistence approach which writes in smaller batches and updates the epoch marker only after the full epoch is populated. Unfortunately that pattern doesn't apply here since we're updating values in place rather than creating new entries. Merging.

PR #254, opened 2025-10-16
Allow offline compilation and `--config <PATH>`

michalrus:
### Offline compilation

* You reference a JSON file with the `ACROPOLIS_OFFLINE_MIRROR` environment variable. 
* Each key is an URL that the `build.rs` scripts would normally download. 
* Each value is its local path on disk. 
* If the variable or keys are not set, it downloads as usual. But people compiling without Internet access, can still do so by providing their own pre-downloaded overrides.

### Specifying `--config <FILE>`

* You can now pass the executables a `--config <FILE>` argument, and that file will be used.
* If you set `ACROPOLIS_OMNIBUS_DEFAULT_CONFIG` during compilation, this value will be used as the default, allowing build systems to override the default default of `omnibus.toml`.
* Same for `replayer.toml` and `ACROPOLIS_REPLAYER_DEFAULT_CONFIG`.

```
❯ acropolis_process_omnibus --help
Usage: acropolis_process_omnibus [OPTIONS]

Options:
      --config <PATH>  [default: /nix/store/3jnims9fplpmbjl4lacq2bny0xqxfdwm-omnibus.toml]
  -h, --help           Print help
```

```
❯ acropolis_process_replayer --help
Usage: acropolis_process_replayer [OPTIONS] <--governance-collect|--governance-replay|--alonzo-governance-collect>

Options:
      --config <PATH>              [default: /nix/store/ck6wnix10y6v5n6hr24b1qq18y4lnxjj-replayer.toml]
      --governance-collect
      --governance-replay
      --alonzo-governance-collect
  -h, --help                       Print help
```

### DIFF
--- a/modules/genesis_bootstrapper/build.rs
+++ b/modules/genesis_bootstrapper/build.rs
@@ -15,8 +15,10 @@ async fn fetch_bytes(client: &reqwest::Client, url: &str) -> Result<Vec<u8>> {
     if let Ok(path) = env::var("ACROPOLIS_OFFLINE_MIRROR") {
         if let Ok(file) = File::open(&path) {
             if let Ok(map) = from_reader::<_, HashMap<String, String>>(file) {
-                if let Some(s) = map.get(url) {
-                    return Ok(s.as_bytes().to_vec());
+                if let Some(path) = map.get(url) {
+                    if let Ok(bytes) = fs::read(&Path::new(path).to_path_buf()) {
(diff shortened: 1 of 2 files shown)


michalrus:
> JSON file to reference file paths, and not their contents as JSON strings?

Done in 976793e75c47334176f83ad6f4fa7e6213c6ccbb.

PR #261, opened 2025-10-20
SPDD validation test

whankinsiv:
This PR adds a validation tool to compare the Acropolis-generated SPDD with the epoch stakes reported by DB Sync. 
It provides a trusted reference for verifying our active-stake snapshot and helps identify discrepancies between the Acropolis state and the real on-chain data. Done as part of #234. 

### Bugs found during validation:
* MIRs not counted towards active stake in epoch 208. 
Fix: Pay MIRs to reward account at the start of the epoch instead of delaying until reward calculation. 
* Retired pools not removed from SPDD 
Fix: Filter based on active pool set.  
* Retired pools not receiving their deposit back at the correct boundary 
Note: I was unable to find the fix but this may be an early Shelley era bug. In epoch 209 we have 3 pools missing 500 ADA of active stake due to deregistration and subsequent reregistration of their pools.

### Example output 
(The 42 pool mismatch is due to MIRs not being included.)
Screenshot 2025-10-20 at 2 31 11 PM

### Other changes:
* Refactored the `spdd_retation_epochs` config option in `omnibus.toml` to always require an integer, removing the separate `store-spdd-history` flag (Set to 0 by default which is equivalent to `store-spdd-history=false`). 
* Re-added the `/spdd?epoch={n}` REST handler in `spdd_state`, which is used by the validator. 
* Fixed elided lifetime warnings for `StakeAddressMap` methods.

### DIFF
--- a/modules/accounts_state/src/accounts_state.rs
+++ b/modules/accounts_state/src/accounts_state.rs
@@ -8,6 +8,7 @@ use acropolis_common::{
     BlockInfo, BlockStatus,
 };
 use anyhow::Result;
+use bigdecimal::Zero;
 use caryatid_sdk::{message_bus::Subscription, module, Context, Module};
 use config::Config;
 use std::sync::Arc;
@@ -438,15 +439,6 @@ impl AccountsState {
             .max(0) as u64;
         info!("SPDD retention epochs: {:?}", spdd_retention_epochs);
 
-        if spdd_retention_epochs > 0 {
-            info!("SPDD database path: {}", spdd_db_path);
-            info!(
-                "SPDD retention: {} epochs (~{} GB max)",
-                spdd_retention_epochs,
-                (spdd_retention_epochs as f64 * 0.12).ceil()
-            )
-        }
-
         // Query topics
         let accounts_query_topic = config
             .get_string(DEFAULT_ACCOUNTS_QUERY_TOPIC.0)
@@ -475,7 +467,7 @@ impl AccountsState {
         let history_tick = history.clone();
 
         // Spdd store
-        let spdd_store = if spdd_retention_epochs > 0 {
+        let spdd_store = if !spdd_retention_epochs.is_zero() {
             Some(Arc::new(Mutex::new(SPDDStore::load(
                 std::path::Path::new(&spdd_db_path),
                 spdd_retention_epochs,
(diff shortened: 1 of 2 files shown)


modules/accounts_state/src/accounts_state.rs
```diff
@@ -484,7 +475,7 @@ impl AccountsState {
         let history_tick = history.clone();
 
         // Spdd store
-        let spdd_store = if store_spdd_history {
+        let spdd_store = if spdd_retention_epochs > 0 {
```
lowhung: If this logic is coupled for logging + creating this new arc maybe put it in a function so we have one source of truth for that?

modules/accounts_state/src/spo_distribution_store.rs
```diff
@@ -137,11 +131,9 @@ impl SPDDStore {
         let marker_key = encode_epoch_marker(epoch);
         self.epoch_markers.insert(marker_key, b"complete")?;
 
-        if let Some(retention) = self.retention_epochs {
-            if epoch >= retention {
-                let keep_from_epoch = epoch - retention + 1;
-                self.prune_epochs_before(keep_from_epoch)?;
-            }
+        if epoch >= self.retention_epochs {
+            let keep_from_epoch = epoch - self.retention_epochs + 1;
```
lowhung: Nice 👍

modules/accounts_state/src/accounts_state.rs
lowhung: ^Mostly a nit

golddydev (approved):

modules/spdd_state/src/rest.rs
```diff
@@ -0,0 +1,55 @@
...
+                    404,
+                    &format!("SPDD not found for epoch {}", epoch),
+                ));
+            }
+        },
+        None => locked.get_latest(),
+    };
+
+    if let Some(spdd) = spdd_opt {
+        let spdd: HashMap<_, _> = spdd
```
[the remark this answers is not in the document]
lowhung: I know the rustc can explicitly set the type, but when it's not clear what the result of the map function returns it can be useful to explicitly write it out for readability sake. But if it's really obvious I think it's fair to just leave it.

sandtreader:
Very happy with this :-)  Interested to know if there was a reason you wrote the test in TS - ease of CI?

Issue #227, opened 2025-10-02
Consensus Validation Coordination / Flight Control


buddhisthead:
Implementation of the go/no-go decision for a chain. Broke out from #201

The implementation of each module is in a separate ticket.
Design is in #228

Issue #205, opened 2025-09-29
Multi-peer with consensus


sandtreader:
Fork UpstreamChainFetcher into PeerNetworkInterface, and pass all ChainSync results to the Consensus module (see #201), which will then indicate which blocks to fetch based on its chain selection.

Note this is still manually configured with a topology file.

Issue #197, opened 2025-09-29
VRF validation


sandtreader:
Validate the VRF calculation in the block header, based on the slot number, the quoted VRF key hash and the SPDD you can look up the SPO to find the proportion of stake.  There should be support for this in Pallas real soon now - check with Pavo.

On either success or failure it should generate a ValidationResult message (TBD) on "cardano.validation.vrf".

buddhisthead:
how is this ticket going? Does it feel like it will close by the end of this sprint? Thanks.

PR #262, opened 2025-10-21
227 consensus validation / Flight Control

sandtreader:
Resolves issue #227 

This PR implements the skeleton of the "Flight Control" Go/NoGo mechanism for validation:

- Merges BlockHeader and BlockBody messages into RawBlock, switches handling in EpochState, SPOState
- Definition of BlockValidation messages which contain a ValidationResult (in common::validation)
- New consensus module which interposes in the (merged) block message stream, accepting `cardano.block.available` from Mithril/Upstream and sending on `cardano.block.proposed` (which downstreams all now switch to)
- Configuration of validators in consensus - currently empty because there aren't any.  VRF Validator (#197) can be the first one.

For now consensus can't do anything with a validation failure other than log it, because it doesn't have multiple chains to choose from. #205 will extend this.

I've requested lots of reviewers because (a) this is pretty central to a lot we are doing, and comments welcome and (b) almost everyone will be involved in implementing validators which need to connect to it.

### DIFF
--- a/processes/omnibus/src/main.rs
+++ b/processes/omnibus/src/main.rs
@@ -61,11 +61,9 @@ pub async fn main() -> Result<()> {
     let args = <self::Args as clap::Parser>::parse();
 
     // Standard logging using RUST_LOG for log levels default to INFO for events only
-    let fmt_layer = fmt::layer().with_filter(EnvFilter::from_default_env());
-
-    // TODO disabled this filter because it prevents debugging - investigate
-    //.add_directive(filter::LevelFilter::INFO.into()))
-    //        .with_filter(filter::filter_fn(|meta| meta.is_event()));
+    let fmt_layer = fmt::layer()
+        .with_filter(EnvFilter::from_default_env().add_directive(filter::LevelFilter::INFO.into()))
+        .with_filter(filter::filter_fn(|meta| meta.is_event()));
 
     // Only turn on tracing if some OTEL environment variables exist
     if std::env::vars().any(|(name, _)| name.starts_with("OTEL_")) {


modules/consensus/src/consensus.rs
```diff
@@ -0,0 +1,124 @@
...
+                    return;
+                };
+                match message.as_ref() {
+                    Message::Cardano((block_info, CardanoMessage::BlockAvailable(_block_msg))) => {
+                        let span = info_span!("consensus", block = block_info.number);
+
+                        async {
+                            // TODO Actually decide on favoured chain!
+
+                            // Send to downstreams to validate
```
buddhisthead: This comment sounds like we're sending the blocks to other nodes, but it's to the validators, right?

modules/consensus/src/consensus.rs
```diff
@@ -0,0 +1,124 @@
...
+                            let results: Vec<_> =
+                                try_join_all(validator_subscriptions.iter_mut().map(|s| s.read()))
+                                    .await
+                                    .unwrap_or_else(|e| {
+                                        error!("Failed to read validations: {e}");
+                                        vec![]
+                                    });
+
+                            // All must be positive!
+                            let all_say_go = results.iter().fold(true, |all_ok, (_topic, msg)| {
```
buddhisthead: It's always been my preference to use a term like `accum` for the fold accumulator so it's easy to read. But I can get used to this!

modules/consensus/src/consensus.rs
```diff
@@ -0,0 +1,124 @@
...
+                            // Send to downstreams to validate
+                            context
+                                .message_bus
+                                .publish(&publish_blocks_topic, message.clone())
+                                .await
+                                .unwrap_or_else(|e| error!("Failed to publish: {e}"));
+
+                            // Read validation responses from all validators in parallel
+                            let results: Vec<_> =
+                                try_join_all(validator_subscriptions.iter_mut().map(|s| s.read()))
```
buddhisthead: Does this deal with a validator that doesn't respond, like a timeout?

modules/consensus/src/consensus.rs
sandtreader: Yeah, that's confusing, I'll change the comment

modules/consensus/src/consensus.rs
sandtreader: It didn't feel like we were accumulating...  TBH I would have written this as a loop until recently ;-)

modules/consensus/src/consensus.rs
sandtreader: Yes that's right - either of which can flag a failure.

processes/omnibus/src/main.rs
```diff
@@ -59,9 +60,11 @@ pub async fn main() -> Result<()> {
     let args = <self::Args as clap::Parser>::parse();
 
     // Standard logging using RUST_LOG for log levels default to INFO for events only
-    let fmt_layer = fmt::layer()
-        .with_filter(EnvFilter::from_default_env().add_directive(filter::LevelFilter::INFO.into()))
-        .with_filter(filter::filter_fn(|meta| meta.is_event()));
+    let fmt_layer = fmt::layer().with_filter(EnvFilter::from_default_env());
+
+    // TODO disabled this filter because it prevents debugging - investigate
```
alexwoods: Was it intentional to check this in?

modules/consensus/src/consensus.rs
```diff
@@ -0,0 +1,124 @@
...
+                                    .await
+                                    .unwrap_or_else(|e| {
+                                        error!("Failed to read validations: {e}");
+                                        vec![]
+                                    });
+
+                            // All must be positive!
+                            let all_say_go = results.iter().fold(true, |all_ok, (_topic, msg)| {
```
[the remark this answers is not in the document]
sandtreader: True, I'll simplify

Issue #163, opened 2025-09-10
Replace RewardAccount with StakeAddress


sandtreader:
The definition of PoolRegistration and GovernanceProposal both use RewardAccount (which is a Vec<u8>) to hold the account that deposits should be returned to.  This is held as the binary encoding of the address, with the 'e1' header (usually - I'm not sure whether script addresses can be used as well?).

This means there are several places where we need to extract the hash to match with other uses (e.g. in AccountState), which means reconstructing the StakeAddress from binary and then using get_hash().

It would be better if we stored the accounts as StakeAddress to begin with, either by redefining RewardAccount or replacing it.

PR #265, opened 2025-10-22
Add `NetworkId` into `BlockInfo`

lowhung:
### Description

Resolves issue #163 

This adds `NetworkId` into the `BlockInfo` struct via the Mithril and Upstream fetchers, and introduces the identifier into the account state, drep state, epochs state and rest_blockfrost as these all rely on querying the `StakeAddressMap` which has had it's API updated to require `StakeAddress` structs as its key.

### DIFF
--- a/common/src/state_history.rs
+++ b/common/src/state_history.rs
@@ -54,7 +54,7 @@ impl<S: Clone + Default> StateHistory<S> {
 
     /// Get the current state assuming any rollback has been done
     /// Cloned for modification - call commit() when done
-    pub fn  get_current_state(&self) -> S {
+    pub fn get_current_state(&self) -> S {
         self.history.back().map(|entry| entry.state.clone()).unwrap_or_default()
     }
 
--- a/common/src/types.rs
+++ b/common/src/types.rs
@@ -6,7 +6,6 @@ use crate::{
     address::{Address, ShelleyAddress, StakeAddress},
     protocol_params,
     rational_number::RationalNumber,
-    AddressNetwork, StakeAddressPayload,
 };
 use anyhow::{anyhow, bail, Error, Result};
 use bech32::{Bech32, Hrp};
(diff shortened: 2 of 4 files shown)


modules/mithril_snapshot_fetcher/src/mithril_snapshot_fetcher.rs
```diff
@@ -307,6 +308,12 @@ impl MithrilSnapshotFetcher {
 
                         let timestamp = genesis.slot_to_timestamp(slot);
 
+                        let network_id = NetworkId::from(
+                            config
+                                .get_string("network-id")
+                                .unwrap_or(DEFAULT_NETWORK_ID.to_string()),
+                        );
+
```
sandtreader: Yes great

sandtreader (changes requested):
Main issue is not to store networkId in state - there's no need, and no sensible place to capture it (first message?  Any message?).  It should always be available in BlockInfo except for things that create blocks where it's configured (as you have done).

To avoid having to pass it around I've suggested changing parameters to be full StakeAddresses, so the caller (which is usually a message handler) can do the conversion.  Even better, the messages should have StakeAddresses instead of StakeCredentials (which may fall away altogether), so it's things like TxUnpacker which do the conversion based on the NetworkID they get in their BlockInfo.

If we apply the same change to the REST query responses I think we can avoid having to configure network in RESTBlockfrost too.

common/src/address.rs
```diff
@@ -28,6 +26,15 @@ pub enum AddressNetwork {
...
 
+impl From<NetworkId> for AddressNetwork {
+    fn from(network: NetworkId) -> Self {
+        match network {
+            NetworkId::Mainnet => Self::Main,
+            NetworkId::Testnet => Self::Test,
+        }
+    }
+}
+
```
whankinsiv: Why is `AddressNetwork`?

Issue #260, opened 2025-10-20
CI/CD for basic Rust hygiene


buddhisthead:
Setup a CI/CD pipeline to gate PRs on:
- clippy
- fmt
- no warnings
- audit stuff

PR #274, opened 2025-10-23
chore: clippify acropolis_common

SupernaviX:
Partially addresses #260 

Begin a _gradual_ adoption of clippy throughout the codebase. I'm planning to do this one package at a time, to avoid gigantic PRs and neverending merge conflicts. I'm also starting from the shared libraries, because clippy would report errors from them when linting any package which depended on them.


sandtreader:
Wow, good stuff!  Is clippy doing any of this automatically or have you just been climbing a wall of errors?

PR #287, opened 2025-10-24
Refactor: Replace `StakeAddressPayload` with `StakeCredential`

lowhung:


common/src/types.rs
```diff
@@ -625,15 +625,16 @@ pub struct PotDelta {
...
 )]
 pub enum Credential {
     /// Script hash. NOTE: Order matters when parsing Haskell Node Snapshot data.
-    ScriptHash(KeyHash),
+    ScriptHash(#[serde_as(as = "Hex")] KeyHash),
 
     /// Address key hash
-    AddrKeyHash(KeyHash),
+    AddrKeyHash(#[serde_as(as = "Hex")] KeyHash),
 }
```
whankinsiv: Credential should use fixed size hashes using the `Hash` type that Chris has added instead of KeyHash. It looks like he has already defined `ScriptHash` and `AddrKeyhash` in `/common/src/hash.rs`.

whankinsiv (approved):
Please create an issue for switching Credential to use the types defined in hash.rs. As discussed this is a larger refactor which can be addressed in a follow up PR due to all the `to_vec()` conversions in the codebase.

Issue #256, opened 2025-10-16
Implement BlockFrost /accounts endpoints


sandtreader:
We need to implement the /accounts endpoint, which requires storage of history of changes in accounts.  Create a new module (AccountsHistory?) to avoid complicating AccountsState any more.  Anything that can be got from the current state can still come from AccountsState.

https://docs.blockfrost.io/#tag/cardano--accounts

PR #292, opened 2025-10-24
feat: registration, delegation, and MIR history account endpoints

whankinsiv:
This PR implements 3 of the 10 remaining account related endpoints described in #256:
* `/accounts/{stake_address}/registrations` - Queries `historical_accounts_state` for registration history, then fetches `TxHash` values from `chain_store` using each `TxIdentifier` to construct the Blockfrost aligned response. 
* `/accounts/{stake_address}/delegations` - Queries `historical_accounts_state` for delegation history, resolving `TxHash` via `chain_store`. 
* `/accounts/{stake_address}/mirs` - Queries `historical_accounts_state` for MIR history, resolving `TxHash` via `chain_store`. 

Main changes:
* Registered 3 endpoints in `rest_blockfrost` initialization. 
* Added and implemented `handle_account_registrations_blockfrost`, `handle_account_delegations_blockfrost`, and `handle_account_mirs_blockfrost` to `handlers/accounts.rs`. 

Notes:
* The registrations endpoint currently returns 0 for the amount field as this data isn't included in the certificate message. I am still considering the best approach for handling this and am currently leaning towards moving the historical epoch stakes data from `accounts_state` to `historical_accounts_state` although this would require passing the per stake address distribution on the message bus. Alternatively, we could rely on a second query in the rest handler to retrieve the account epoch stake from `accounts_state` under the assumption that `spdd-retention-epochs` is configured to store all epochs. 

Next steps:
* The next PR for #256 will add withdrawal processing to `historical_accounts_state` and implement the corresponding REST handler.

### DIFF
--- a/modules/rest_blockfrost/src/handlers/accounts.rs
+++ b/modules/rest_blockfrost/src/handlers/accounts.rs
@@ -50,27 +50,25 @@ pub async fn handle_single_account_blockfrost(
         |message| match message {
             Message::StateQueryResponse(StateQueryResponse::Accounts(
                 AccountsStateQueryResponse::AccountInfo(account),
-            )) => Ok(account),
+            )) => Ok(Some(account)),
             Message::StateQueryResponse(StateQueryResponse::Accounts(
                 AccountsStateQueryResponse::NotFound,
-            )) => {
-                return Err(anyhow::anyhow!("Account not found"));
-            }
+            )) => Ok(None),
             Message::StateQueryResponse(StateQueryResponse::Accounts(
                 AccountsStateQueryResponse::Error(e),
-            )) => {
-                return Err(anyhow::anyhow!(
-                    "Internal server error while retrieving account info: {e}"
-                ));
-            }
-            _ => {
-                return Err(anyhow::anyhow!(
-                    "Unexpected message type while retrieving account info"
-                ))
-            }
+            )) => Err(anyhow::anyhow!(
+                "Internal server error while retrieving account info: {e}"
+            )),
+            _ => Err(anyhow::anyhow!(
+                "Unexpected message type while retrieving account info"
+            )),
         },
     )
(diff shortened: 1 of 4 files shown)


modules/rest_blockfrost/src/handlers/accounts.rs
```diff
@@ -115,15 +105,332 @@ pub async fn handle_single_account_blockfrost(
...
+                "Missing tx hash for delegation",
+            ));
+        };
+
+        let pool_id = match r.pool.to_bech32_with_hrp("pool") {
+            Ok(p) => p,
+            Err(e) => {
+                return Ok(RESTResponse::with_text(
+                    500,
+                    &format!("Failed to encode pool ID: {e}"),
```
SupernaviX: This is a bigger change than I think makes sense in this PR, but in other projects I usually implement an API error to deal with the boilerplate of friendly REST error messages:

```rs
pub struct AppError(u16, anyhow::Error);
impl From<AppError> for RESTResponse {
    fn from(value: AppError) -> Self {
        Self::with_text(value.0, value.1.to_string())
    }
}
impl<E> From<E> for AppError
where
    E: Into<anyhow::Error>,
{
    fn from(value: E) -> Self {
        Self(500, value.into())
    }
}
```

It lets you lean on anyhow, and write calls to fallible methods in a much briefer way:
```rs
let pool_id = r.pool.to_bech32_with_hrp("pool").context("Failed to encode pool id")?;
```

SupernaviX (approved):

modules/rest_blockfrost/src/handlers/accounts.rs
whankinsiv: I've created #313 so that this refactor can be implemented in a follow up PR. This will clean up the module significantly.

Issue #257, opened 2025-10-16
Replace direct KeyHash usage with typed wrappers


whankinsiv:
`KeyHash` is currently used in multiple contexts to represent different entities: 
* spo_id
* pool_operator
* reward_account

This reduces clarity and type safety in the codebase. Create wrapper types for these fields and refactor existing code to use them.

PR #321, opened 2025-10-29
Refactor: Combine byte array macros with hash.rs

lowhung:
### What this PR does

Implements #257

This PR combines the byte array macros @alexwoods wrote with the hash declarations so we get both options in one file so it's easier to find.

### DIFF
--- a/modules/tx_unpacker/src/tx_unpacker.rs
+++ b/modules/tx_unpacker/src/tx_unpacker.rs
@@ -342,7 +342,7 @@ impl TxUnpacker {
                                         if publish_governance_procedures_topic.is_some() {
                                             if let Some(pp) = props {
                                                 // Nonempty set -- governance_message.proposal_procedures will not be empty
-                                                let mut proc_id = GovActionId { transaction_id: TxHash::from(*tx.hash()), action_index: 0 };
+                                                let mut proc_id = GovActionId { transaction_id: tx_hash, action_index: 0 };
                                                 for (action_index, pallas_governance_proposals) in pp.iter().enumerate() {
                                                     match proc_id.set_action_index(action_index)
                                                         .and_then (|proc_id| map_parameters::map_governance_proposals_procedures(proc_id, pallas_governance_proposals))
@@ -356,7 +356,7 @@ impl TxUnpacker {
                                             if let Some(pallas_vp) = votes {
                                                 // Nonempty set -- governance_message.voting_procedures will not be empty
                                                 match map_parameters::map_all_governance_voting_procedures(pallas_vp) {
-                                                    Ok(vp) => voting_procedures.push((TxHash::from(*tx.hash()), vp)),
+                                                    Ok(vp) => voting_procedures.push((tx_hash, vp)),
                                                     Err(e) => error!("Cannot decode governance voting procedures in slot {}: {e}", block.slot)
                                                 }
                                             }


common/src/hash.rs
```diff
@@ -46,11 +64,33 @@ impl<const BYTES: usize> From<[u8; BYTES]> for Hash<BYTES> {
...
 
-impl<const BYTES: usize> From<&[u8]> for Hash<BYTES> {
-    fn from(value: &[u8]) -> Self {
-        let mut hash = [0; BYTES];
-        hash.copy_from_slice(value);
-        Self::new(hash)
+impl<const BYTES: usize> TryFrom<&[u8]> for Hash<BYTES> {
+    type Error = std::array::TryFromSliceError;
+
+    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
```
lowhung: I think `try_from` is more appropriate than allowing a `from` if we don't know the size of the array we're working with. Thoughts?

whankinsiv: This seems logical.

common/src/hash.rs
```diff
@@ -46,11 +64,33 @@ impl<const BYTES: usize> From<[u8; BYTES]> for Hash<BYTES> {
...
+        Ok(Self::new(hash))
+    }
+}
+
+impl<const BYTES: usize> TryFrom<Vec<u8>> for Hash<BYTES> {
+    type Error = Vec<u8>;
+
+    fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
+        let hash: [u8; BYTES] = value.try_into()?;
+        Ok(Self::new(hash))
```
whankinsiv: This can be `Self::try_from(value.as_slice())`

common/src/hash.rs
```diff
@@ -125,6 +178,152 @@ impl<'a, C, const BYTES: usize> minicbor::Decode<'a, C> for Hash<BYTES> {
...
+        }
+
+        impl From<[u8; $size]> for $name {
+            fn from(bytes: [u8; $size]) -> Self {
+                Self(Hash::new(bytes))
+            }
+        }
+
+        impl TryFrom<Vec<u8>> for $name {
+            type Error = Vec<u8>;
```
whankinsiv: This should return context of why TryFrom failed. Either an `anyhow::Error` or `String`.

common/src/hash.rs
```diff
@@ -125,6 +178,152 @@ impl<'a, C, const BYTES: usize> minicbor::Decode<'a, C> for Hash<BYTES> {
...
+                ctx: &mut C,
+            ) -> Result<Self, minicbor::decode::Error> {
+                Ok(Self(Hash::decode(d, ctx)?))
+            }
+        }
+
+        impl crate::serialization::Bech32Conversion for $name {
+            fn to_bech32(&self) -> Result<String, anyhow::Error> {
+                use crate::serialization::Bech32WithHrp;
+                self.0.to_vec().to_bech32_with_hrp($hrp)
```
whankinsiv: This could be `self.as_ref().to_bech32_with_hrp($hrp)` to avoid the to_vec() allocation.

common/src/hash.rs
lowhung: Made the update here in this commit e00ad781bb29c97622a704dd7eb2b5ec366b89d7

common/src/hash.rs
lowhung: The `as_ref()` in this case is returning a `&[u8]` so I don't think this will work. The implementation of the to_bech is on the vector.

```
    |
314 |                 self.0.as_ref().to_bech32_with_hrp($hrp).with_context(|| {
    |                                 ^^^^^^^^^^^^^^^^^^ method not found in `&[u8]`
    |
   ::: common/src/types.rs:457:1
    |
457 | declare_hash_type_with_bech32!(VrfKeyHash, 32, "vrf_vk");
    | -------------------------------------------------------- in this macro invocation
```

common/src/hash.rs
```diff
@@ -46,11 +64,33 @@ impl<const BYTES: usize> From<[u8; BYTES]> for Hash<BYTES> {
...
+impl<const BYTES: usize> TryFrom<Vec<u8>> for Hash<BYTES> {
+    type Error = Vec<u8>;
+
+    fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
+        let hash: [u8; BYTES] = value.try_into()?;
+        Ok(Self::new(hash))
+    }
+}
+
+impl<const BYTES: usize> From<Hash<BYTES>> for Vec<u8> {
```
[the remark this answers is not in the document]
lowhung: Made the update here in this commit e00ad781bb29c97622a704dd7eb2b5ec366b89d7

common/src/hash.rs
```diff
@@ -125,6 +178,152 @@ impl<'a, C, const BYTES: usize> minicbor::Decode<'a, C> for Hash<BYTES> {
...
+            fn to_bech32(&self) -> Result<String, anyhow::Error> {
+                use crate::serialization::Bech32WithHrp;
+                self.0.to_vec().to_bech32_with_hrp($hrp)
+            }
+
+            fn from_bech32(s: &str) -> Result<Self, anyhow::Error> {
+                use crate::serialization::Bech32WithHrp;
+                let v = Vec::<u8>::from_bech32_with_hrp(s, $hrp)?;
+                Self::try_from(v).map_err(|_| {
+                    anyhow::Error::msg(format!("Bad vector input to {}", stringify!($name)))
```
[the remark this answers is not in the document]
lowhung: Made the update here in this commit e00ad781bb29c97622a704dd7eb2b5ec366b89d7

common/src/hash.rs
lowhung: Done here acbeb4617d28f23e445551e9d843bea24796cee3

modules/tx_unpacker/src/tx_unpacker.rs
```diff
@@ -183,7 +183,7 @@ impl TxUnpacker {
                                             for input in inputs {  // MultiEraInput
                                                 // Lookup and remove UTxOIdentifier from registry 
                                                 let oref = input.output_ref();
-                                                let tx_ref = TxOutRef::new(TxHash(**oref.hash()), oref.index() as u16);
+                                                let tx_ref = TxOutRef::new(TxHash::from(**oref.hash()), oref.index() as u16);
```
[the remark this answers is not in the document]
lowhung: Nice catch there, I updated to use the converted hash on 166
21f5388e233b6c9dc09fecec905bc4bcf8d6a9cd

PR #327, opened 2025-10-30
feat: historical accounts Shelley address indexing

whankinsiv:
This PR implements `ShelleyAddress` indexing by `StakeAddress` in `historical_accounts_state` along with the corresponding REST handler as part of #256. Addresses are returned in the order they were first discovered on chain to match Blockfrost's return schema. 

### Main changes
* Refactored `stake_delta_filter` to group deltas by stake address and added a new `addresses` field to `StakeAddressDelta`, providing context on which Shelley addresses were involved in each account delta. 
* Updated `mithril_snapshot_fetcher` to include time deltas between pauses, providing a loose benchmarking mechanism for different implementations.
* Implemented placeholder methods `handle_address_deltas`, `get_addresses`, and `GetAccountAssociatedAddresses` query handler in `historical_accounts_state`
* Added and registered the `/accounts/{stake_address}/addresses` endpoint via `handle_account_addresses_blockfrost`. 

### Performance evaluation
3 persistence approaches were evaluated:
1. Vec read/append/write: 
   Encoded Vec approach was extremely inefficient, consuming ~7x the storage of the duplicated key strategy. 
2. read/filter/write new keys: 
   Acceptable performance but ~10% slower due to read before write checks. 
   * Epoch 500: 1.4 GB DB size, 75.7 minute sync
3. Blind duplicate writes (implemented): 
   Writes new keys without reads with filtering handled in the getter. 
   * Epoch 500: 2.1GB DB size, 68.7 minute sync

### Next steps:
The next PR will implement:
* `/accounts/{stake_address}/addresses/total`
* `/acounts/{stake_address}/addresses/assets`
* `/accounts/{stake_address}/utxos`

### DIFF
--- a/common/src/address.rs
+++ b/common/src/address.rs
@@ -353,6 +353,11 @@ impl ShelleyAddress {
         }
 
         let header = data[0];
+
+        if header & 0x80 != 0 {
+            return Err(anyhow!("invalid header: high bit set"));
+        }
+
         let network = match header & 0x0F {
             0 => NetworkId::Testnet,
             1 => NetworkId::Mainnet,


processes/omnibus/omnibus.toml
```diff
@@ -71,7 +71,7 @@ store-registration-history = false
 store-delegation-history = false
 store-mir-history = false
 store-withdrawal-history = false
-store-addresses = false
+store-addresses = true
```
lowhung: Should this be `true`?

lowhung (approved):

modules/mithril_snapshot_fetcher/src/mithril_snapshot_fetcher.rs
```diff
@@ -433,9 +430,9 @@ impl MithrilSnapshotFetcher {
 }
 
 /// Async helper to prompt user for pause behavior
-async fn prompt_pause(description: String, next_description: String) -> bool {
+async fn prompt_pause(description: String) -> bool {
     info!(
-        "Paused at {description}. Press [Enter] to step to {next_description}, or [c + Enter] to continue without pauses."
```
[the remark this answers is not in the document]
whankinsiv: This was removed because next description needed to be called after incrementing to the next pause value. We don't want to increment the pause value (and the start time) until after the user ends the pause.

processes/omnibus/omnibus.toml
whankinsiv: Oups, thanks for catching this. Set to false in 5f6625a.

common/src/address.rs
```diff
@@ -341,7 +342,69 @@ impl ShelleyAddress {
...
+
+        let header = data[0];
+        let network = match header & 0x0F {
+            0 => NetworkId::Testnet,
+            1 => NetworkId::Mainnet,
+            _ => return Err(anyhow!("invalid network bits in header")),
+        };
+
+        let payment_bits = (header >> 4) & 0x01;
+        let delegation_bits = (header >> 5) & 0x03;
```
SupernaviX:
```suggestion
        let delegation_bits = (header >> 5) & 0x07;
```
Looking at CIP-0019, there are 8 valid types of Shelley addresses (2 kinds of payment part, 4 kinds of delegation part). The high bit of the header should always be 0, and if it isn't then this isn't a valid Shelley address.

This change will make the method reject header types 8 through 15 with an "invalid delegation bits" error.

whankinsiv: I've added a guard for `header & 0x80 != 0` in 39a8919.

PR #333, opened 2025-11-03
Snapshot parser with TODO's for handling specific state modules

buddhisthead:
This PR updates the snapshot_bootstrapper module to read and parse an Amaru/Haskell node snapshot file, providing the initial structure of the process. It includes the parser and placeholder callback implementation for the streaming snapshot parser and also the pub/sub top-level message sequence.

Some features:

    Progress Tracking: Logs every million UTXOs processed
    Comprehensive Logging: Detailed statistics and metadata logging
    Error Handling: Proper Result types and error propagation
    Async Architecture: Non-blocking operation with proper context management
    Message Publishing: Publishes GenesisComplete message to trigger next phase

TESTING
cargo test shouldn't be broken, but I have tried in vein to create a mini snapshot file that is 5MB or less. It seems not to be possible (after several hours of trying). I can use the integration test, but I can't think of a unit test yet.

### DIFF
--- a/modules/snapshot_bootstrapper/src/snapshot_bootstrapper.rs
+++ b/modules/snapshot_bootstrapper/src/snapshot_bootstrapper.rs
@@ -221,6 +221,7 @@ impl SnapshotCallbacks for SnapshotHandler {
 
 impl SnapshotBootstrapper {
     pub async fn init(&self, context: Arc<Context<Message>>, config: Arc<Config>) -> Result<()> {
+        // TODO: read a config file path, not the snapshot-path; implement TODOs below.
         let file_path = config
             .get_string("snapshot-path")
             .inspect_err(|e| error!("failed to find snapshot-path config: {e}"))?;
@@ -243,6 +244,12 @@ impl SnapshotBootstrapper {
                 return;
             };
             info!("Received startup message");
+
+            // TODO:
+            // Read config file per docs in NOTES.md
+            // read nonces
+            // read headers
+            // read and process ALL of the snapshot files, not just one.
 
             let span = info_span!("snapshot_bootstrapper.handle");
             async {
--- a/common/src/snapshot/NOTES.md
+++ b/common/src/snapshot/NOTES.md
@@ -30,6 +30,8 @@ There is a path for each network bootstrap configuration file. Network Should
 be one of 'mainnet', 'preprod', 'preview' or 'testnet_<magic>' where
 `magic` is a 32-bits unsigned value denoting a particular testnet.
 
+Data structure, e.g. as [Amaru mainnet](https://github.com/pragma-org/amaru/tree/main/data/mainnet)
+
 The bootstrapper will be given a path to a directory that is expected to contain
 the following files: snapshots.json, nonces.json, and headers.json. The path will
 be used as a prefix to resolve per-network configuration files


modules/snapshot_bootstrapper/src/snapshot_bootstrapper.rs
```diff
@@ -1,16 +1,41 @@
...
+    snapshot_topic: String,
+
+    // Accumulated data from callbacks
+    metadata: Option<SnapshotMetadata>,
+    utxo_count: u64,
+    pools: Vec<PoolInfo>,
+    accounts: Vec<AccountState>,
+    dreps: Vec<DRepInfo>,
+    proposals: Vec<GovernanceProposal>,
+}
```
whankinsiv: This will also need to include fields for protocol parameters, active_stake snapshots for the previous 2 epochs, and the epoch nonce.

whankinsiv (commented):
This looks good to me. A few things we are missing which I believe will be needed:
* epoch nonce (metadata?)
* protocol parameters
* constitutional committee members
* active stake snapshots for preceding two epochs

modules/snapshot_bootstrapper/src/snapshot_bootstrapper.rs
```diff
@@ -19,23 +44,199 @@ const DEFAULT_STARTUP_TOPIC: &str = "cardano.sequence.start";
...
+        self.context
+            .message_bus
+            .publish(
+                &self.snapshot_topic,
+                Arc::new(Message::Snapshot(
+                    acropolis_common::messages::SnapshotMessage::Startup,
+                )),
+            )
+            .await
+            .map_err(|e| anyhow::anyhow!("Failed to publish start message: {e}"))
```
SupernaviX:
```suggestion
            .context("Failed to publish start message")
```
It's not a huge deal, but using this helper from `anyhow::Context` can get you better error messages. It builds a chain of errors (though callers have to format with `"{e:#}"` or `"{e:?}"` to print the chain), unlike `anyhow!("Failed to publish start message: {e}")` which throws out nested backtraces.

buddhisthead:
> This looks good to me. A few things we are missing which I believe will be needed:
> 
>     * epoch nonce (metadata?)
> 
>     * protocol parameters
> 
>     * constitutional committee members
> 
>     * active stake snapshots for preceding two epochs

This 

> This looks good to me. A few things we are missing which I believe will be needed:
> 
>     * epoch nonce (metadata?)
> 
>     * protocol parameters
> 
>     * constitutional committee members
> 
>     * active stake snapshots for preceding two epochs

I concur with your feedback, that we need a more complete wrapper for the bootstrapping process that includes the steps you refer to. I am not sure if we need the protocol parameters and constitutional committee members but maybe so. I think what's missing here is a design document that outlines the needs. It does require three snapshots, nonces, tip, headers, and possibly more.

How would you feel about accepting this PR as a phase 1 that shows the callback implementations and general start/completion sequence; and then we'll do a design document and implement that as the next two tickets?

modules/snapshot_bootstrapper/src/snapshot_bootstrapper.rs
buddhisthead: I've applied this feedback. Thanks.

lowhung (approved):

Issue #313, opened 2025-10-29
Create anyhow wrapper for unified REST error handling


whankinsiv:
We currently construct error responses manually in `rest_blockfrost` for each failure case (parameter parsing, Bech32 conversion, etc.). As suggested by @SupernaviX, introducing a unified `AppError` type that wraps `anyhow::Error` and implements `From` for `RESTResponse` would simplify error handling across the module. This abstraction will allow REST handlers to use `?` error propagation and ensure consistent error responses.

PR #345, opened 2025-11-06
feat: Integrate `QueryError` into state query responses and refactor REST to match on new NotFound enum

lowhung:
### What

Continues to implement #313 

- Integrates the new `QueryError` into the state modules' state.rs files

### DIFF
--- a/modules/address_state/src/address_state.rs
+++ b/modules/address_state/src/address_state.rs
@@ -210,9 +210,16 @@ impl AddressState {
                     AddressStateQuery::GetAddressUTxOs { address } => {
                         match state.get_address_utxos(address).await {
                             Ok(Some(utxos)) => AddressStateQueryResponse::AddressUTxOs(utxos),
-                            Ok(None) => AddressStateQueryResponse::Error(QueryError::not_found(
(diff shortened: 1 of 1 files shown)


modules/address_state/src/address_state.rs
```diff
@@ -211,21 +210,31 @@ impl AddressState {
                     AddressStateQuery::GetAddressUTxOs { address } => {
                         match state.get_address_utxos(address).await {
                             Ok(Some(utxos)) => AddressStateQueryResponse::AddressUTxOs(utxos),
-                            Ok(None) => AddressStateQueryResponse::NotFound,
-                            Err(e) => AddressStateQueryResponse::Error(e.to_string()),
+                            Ok(None) => AddressStateQueryResponse::Error(QueryError::not_found(
+                                "Address not found",
+                            )),
```
whankinsiv: I noticed in accounts_state you included the account in the not found message but here you don't include the address. Please standardize.

lowhung: Resolved here a77ac5a156201ec47dc49a855dadbdd277e82bcb

Issue #347, opened 2025-11-06
Node runner command


buddhisthead:
### What?

Implement a Makefile entry point at the top-level of the project that will startup and run the Acropolis node.

### Acceptance Criteria

- A user can type `make run` and it will build and start an Acropolis node with a default test network.
- The Makefile starts the node with logging level at "info" but has a make variable that can be over ridden to set it differently

- [ ] Documentation updated
- [ ] How to test written

PR #348, opened 2025-11-06
Update Makefile: add `run` target to run omnibus process

lowhung:
Implements #347


buddhisthead (approved):
Thanks for the update!

Issue #335, opened 2025-11-03
Implement Blockfrost /addresses endpoints


whankinsiv:
We need to implement the REST handlers for the remaining endpoints:
* /addresses/{address}/extended
* /addresses/{address}/total
* /addresses/{address}/utxos
* /addresses/{address}/utxos/{asset}
* /addresses/{address}/transactions

All state processing has already been implemented in `address_state`, so this ticket is scoped only to the REST handler implementations.

Issue #200, opened 2025-09-29
Tx Phase 1 Validation


sandtreader:
Validate transactions in phase 1

(from ChatGPT)

- Well-formed CBOR / basic syntax
  - Transaction decodes correctly from CBOR.
  - All expected fields are present and correctly typed.
- Size limits
  - Transaction size ≤ maxTxSize protocol parameter.
  - Number of inputs, outputs, certificates, witnesses ≤ limits.
- Key validity
  - All input key hashes and witness signatures are valid Ed25519 signatures over the Tx body.
  - VKey witnesses match the inputs they are supposed to authorize.
- Scripts (native or Plutus) referenced are present.
- Fee and value preservation (basic)
  - Declared fee ≥ minFee (protocol parameter).
  - No negative values in outputs.
  - Token bundle policy IDs / asset names well-formed.
-  Timelock validity
   - Transaction validity interval (TTL, start slot, expiry) is syntactically correct.
   - If native scripts with timelock conditions are present, their conditions are checked.
- Auxiliary data structure
   - Metadata, if present, is CBOR-valid and within size limits.

This list should be validated against the spec and Haskell code.

This should output a ValidationResult message on 'cardano.validation.tx-phase-1'.

PR #358, opened 2025-11-13
refactor: group UTxO and address deltas by tx

whankinsiv:
## Description
This PR refactors both `UTXODeltasMessage` and `AddressDeltasMessage` to group deltas per transaction instead of per individual UTxO delta. This fixes incorrect transaction counting in both `address_state` and `historical_accounts_state`, particularly in cases where:
* a transaction has both inputs and outputs to the same address
* multiple addresses belong to the same stake account

Account level transaction counting is now handled separately in `historical_accounts_state`, avoiding the need to index `TxIdentifier` values by address. A new `GetAccountTotalTxCount` query is added to support the `/accounts/{stake_address}/addresses/total` endpoint. 

## Related Issue(s)
* #256 & #335 (prevents tx count inflation)
* #200 (prepares for validating tx input/output balancing)

## How was this tested?
* Updated existing tests to align with per transaction delta grouping.
* Removed obsolete tests after confirming they no longer apply to the redesigned data flow.
* Compared REST results with Blockfrost's `/accounts/{stake_address}/addresses/totals` to confirm that tx counts now match expected behavior. 

## Checklist
- [x] I updated comments
- [x] CI is green for this PR

## Impact / Side effects
* `utxo_state` constructs a temporary `AddressTxMap` per transaction to aggregate UTxO deltas before publishing address level deltas. 
* `StakeDeltasMessage` now includes the per block tx count in the delta.

* Tx deltas accumulation logic in `tx_unpacker/src/tx_unpacker.rs` 
* `handle` in `utxo_state/src/state.rs`
* `process_message` in `stake_delta_filter/src/utils.rs`

### DIFF
--- a/common/src/types.rs
+++ b/common/src/types.rs
@@ -212,7 +212,7 @@ pub struct AddressDelta {
     pub created_utxos: Vec<UTxOIdentifier>,
 
     // Sums of spent and created UTxOs
-    pub sent: ValueDelta,
+    pub sent: Value,
     pub received: Value,
 }
 
@@ -2156,7 +2156,7 @@ pub struct AssetAddressEntry {
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 pub struct TxTotals {
-    pub sent: ValueDelta,
+    pub sent: Value,
     pub received: Value,
 }
 
@@ -2183,7 +2183,7 @@ impl AddAssign for AddressTotals {
 impl AddressTotals {
     pub fn apply_delta(&mut self, delta: &TxTotals) {
         self.received.lovelace += delta.received.lovelace;
-        self.sent.lovelace += delta.sent.lovelace as u64;
+        self.sent.lovelace += delta.sent.lovelace;
 
         for (policy, assets) in &delta.received.assets {
             for asset in assets {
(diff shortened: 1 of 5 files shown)


common/src/types.rs
```diff
@@ -2149,6 +2154,12 @@ pub struct AssetAddressEntry {
     pub quantity: u64,
 }
 
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct TxTotals {
+    pub sent: ValueDelta,
+    pub received: Value,
+}
```
SupernaviX: Why is `sent` a `ValueDelta` but `received` a `Value`?  It looks like in at least a few places, we assume that `sent.lovelace` is positive.

(also, it might make some code a little cleaner if `TxTotals` had an `as_delta` helper method which returned its data as a `ValueDelta`)

modules/historical_accounts_state/src/immutable_historical_account_store.rs
```diff
@@ -297,6 +307,26 @@ impl ImmutableHistoricalAccountStore {
         Ok((!addresses.is_empty()).then_some(addresses))
     }
 
+    pub async fn get_tx_count(&self, account: &StakeAddress) -> Result<Option<u32>> {
+        let mut total_count = 0;
+
+        for result in self.tx_count.prefix(account.get_hash().as_ref()) {
+            let (_, bytes) = result?;
+            let epoch_count = u32::from_le_bytes(bytes[..4].try_into()?);
```
lowhung: French methods 🤣

common/src/types.rs
whankinsiv: Thanks for catching this Simon. I had assumed `ValueDelta` was needed downstream but this isn't the case. I've updated `TxTotals` and `AddressTotals` to use `Value` for the sent field in df09d6f.

SupernaviX (approved):

PR #360, opened 2025-11-13
feat: address UTxOs and transactions REST endpoints

whankinsiv:
## Description
* Adds handlers `/addresses/{address}/utxos` and `/addresses/{address}/transactions` in `rest_blockfrost`. 
* Introduces `GetTransactionHashesAndTimestamps` to avoid multiple `chain_store` queries for timestamps and transaction hashes. 
* Refactors `GetUTxOHashes` to return `QueryError` instead of silently failing when UTxO hashes cannot be retrieved. 
* Refactors `get_utxos` and `get_address_utxos` to return an empty Vec instead of None if an address has ever had UTxOs stored, matching Blockfrost behavior. 
* Adds `clear-on-start` config option to `address_state` with default true. 

## Related Issue(s)
#335 

## How was this tested?
* Compared REST output against Blockfrost for Byron and Shelley addresses across both endpoints. 

## Checklist

- [x] CI is green for this PR

## Impact / Side effects
None

* Query handlers in `chain_store.rs`

### DIFF
--- a/modules/rest_blockfrost/src/handlers/addresses.rs
+++ b/modules/rest_blockfrost/src/handlers/addresses.rs
@@ -14,8 +14,6 @@ use acropolis_common::{
     },
     Address, Value,
 };
-use acropolis_common::{Datum, ReferenceScript};
-use blake2::{Blake2b512, Digest};
 use caryatid_sdk::Context;
 
 /// Handle `/addresses/{address}` Blockfrost-compatible endpoint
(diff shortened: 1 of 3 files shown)


common/src/queries/blocks.rs
```diff
@@ -240,3 +244,9 @@ pub struct UTxOHashes {
     pub block_hashes: Vec<BlockHash>,
     pub tx_hashes: Vec<TxHash>,
 }
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct TransactionHashesAndTimeStamps {
+    pub tx_hashes: Vec<TxHash>,
```
lowhung: Would a map of `timestamp -> tx_hashes` be more appropriate here? And if not, keen to learn :)

modules/rest_blockfrost/src/handlers/addresses.rs
```diff
@@ -166,11 +169,125 @@ pub async fn handle_address_totals_blockfrost(
...
+                };
+                let mut hasher = Blake2b512::new();
+                hasher.update(bytes);
+                let result = hasher.finalize();
+                Some(hex::encode(&result[..32]))
+            }
+            None => None,
+        };
+
+        rest_response.push(UTxOREST {
```
lowhung: What are your thoughts on a constructor (`new(...)`) method for the `UTxOREST` struct that takes the tx hash, block hash etc?

golddydev (approved):

modules/rest_blockfrost/src/handlers/addresses.rs
whankinsiv: Added `UTxOREST::new()` in 56e8801.

Issue #164, opened 2025-09-10
Implement /tx BlockFrost APIs


sandtreader:
Implement the /tx BlockFrost APIs.  This requires ChainStore to provide queries on transactions, for which it will need to unpack blocks and decode the transaction.

There is a lot of transaction decoding / mapping from Pallas in TxUnpacker - this will need to be pulled out into a shared library, but *not* `common`, because we don't want to impose a Pallas dependency on everyone, perhaps 'codec/cardano'?  Ideally the API to this would not require any Pallas types either.

Issue #137, opened 2025-09-02
ChainStore module


sandtreader:
We need an optional module (it's not needed for SPDD) to store historical blocks and Tx for two reasons:

1. To implement the BlockFrost /blocks and /txs APIs, which can query historical block/tx data
2. To act as the block archive once we are a full node and can be asked to synchronise by a client

In the BF APIs, blocks can be looked up by 'latest', hash, number or slot.  Transactions can be looked up by hash only.

The storage layer should be abstracted and made pluggable in the same way as UTXOState.  Fjall (sync version) seems to be the best first candidate.  In-memory doesn't make sense at all, of course.

If storing in a key-value store like Fjall, we need to decide whether to store blocks and derive transactions from them, or store transactions and reconstruct blocks if required.  We must not store both!

For BlockFrost we never regenerate the whole block CBOR so the second seems preferred and simpler.  However when we act as the block archive, we need to be sure we can validly reconstruct the block, which means we also need to store a bunch of other data.  TBD!

SupernaviX:
I have a first pass at the obvious way to build the store: https://github.com/input-output-hk/acropolis/pull/138

There's four partitions:
 - `blocks` has "full" block contents, keyed by hash (need to decide what to actually store)
 - `block-hashes-by-slot` has block hashes, keyed by big-endian slot
 - `block-hashes-by-number` has block hashes, keyed by big-endian number
 - `tx` has tx bodies, keyed by hash

My plan is to add to that store whenever we get a `ReceivedTxs` message

SupernaviX:
Opened #147 for some prereq work (exposing certain computed fields which we need in this API)

sandtreader:
Could we capture the block CBOR header and footer but store references to the Tx bodies somehow, to avoid duplication?

SupernaviX:
I think this is done? the PR is merged

PR #370, opened 2025-11-19
Ajw/164 tx blockfrost api

alexwoods:
## Description

This PR adds some of the txs endpoints to the REST Blockfrost module.

## Related Issue(s)
Relates to #137, #164.

## How was this tested?
Manual comparison against dbsync (mainnet) :-(
Example requests that should work on mainnet:
- info: /txs/95763c53b485088608edfc95a0ef3d512cf5c7727f09e046d80eda0ed437cbe8 (note: utxo info not present - requires historic utxo lookup by hash)
- stakes: /txs/3485af236ba4776a95cefe5d6b4400dff6129c5753521e3774a950766c3c8359/stakes
- delegations: /txs/3485af236ba4776a95cefe5d6b4400dff6129c5753521e3774a950766c3c8359/delegations
- withdrawals: /txs/9efef43f1d8ae696877c29be676055adfcc116510e57439c8a841eb991aa3886/withdrawals
- mirs: /txs/35d2728ea6ad89bf809565c9ed698bb1c5cddf83591ba2e8bba951cb8fee0035/mirs
- pool updates: /txs/841cca81da918feb9fa7257a34630eac95794be712ed3faae6df64f215ce25f2/pool_updates
- pool retirements: /txs/896cf8fefad1eaf0fa056ba3adf28bfb26b06d1beed64cf790deb595dcb2687a/pool_retires
- metadata: /c220e20cc480df9ce7cd871df491d7390c6a004b9252cf20f45fc3c968535b4a/metadata

## Checklist

- [x] CI is green for this PR

## Impact / Side effects
There shouldn't be any side-effects.

I've made custom serialisers within the blockfrost code. I'm unsure if there's a nicer way to do what I was trying to do, which was to keep serialisation out of the common library because in many cases this serialisation is quite peculiar to blockfrost and should not be a default.

### DIFF
--- a/modules/rest_blockfrost/src/handlers/transactions.rs
+++ b/modules/rest_blockfrost/src/handlers/transactions.rs
@@ -88,7 +88,7 @@ impl Serialize for TxOutputAmount {
     }
 }
 
-/// Handle `/txs/{hash}`
+/// Handle `/txs/{hash}[/param][/param2]`
 pub async fn handle_transactions_blockfrost(
     context: Arc<Context<Message>>,
     params: Vec<String>,
@@ -183,6 +183,7 @@ async fn handle_transaction_query(
                 };
                 // TODO: calc from outputs and inputs if recorded_fee is None
                 let fee = txs_info.recorded_fee.unwrap_or_default();
+                // TODO: Check whether updates require a deposit or not
                 let deposit = match calculate_deposit(
                     txs_info.pool_update_count,
                     txs_info.stake_cert_count,


modules/rest_blockfrost/src/handlers/transactions.rs
```diff
@@ -0,0 +1,592 @@
...
+
+impl Serialize for TxPoolRetirementCertificate {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut state = serializer.serialize_struct("TxPoolUpdateCertificate", 3)?;
+        state.serialize_field("cert_index", &self.0.cert_index)?;
+        state.serialize_field("pool_id", &self.0.pool_id.to_string())?;
+        state.serialize_field("retirement_epoch", &self.0.retirement_epoch)?;
```
whankinsiv: `retirement_epoch` should be `retiring_epoch`.

lowhung (approved):

common/src/queries/transactions.rs
```diff
@@ -35,38 +47,159 @@ pub enum TransactionsStateQueryResponse {
...
 
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
-pub struct TransactionInfo {}
+#[derive(Debug, Clone, serde::Deserialize)]
+pub enum TransactionOutputAmount {
+    Lovelace(Lovelace),
+    Asset(NativeAsset),
+}
+
+impl Serialize for TransactionOutputAmount {
```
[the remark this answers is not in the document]
alexwoods: I'm going for a 4th option, which is to apply the standard Serialize derive in `common`, and move the special serialization into `rest_blockfrost`. This is one of the cases where the serialization for blockfrost is kind of unusual due to the numerics being serialized to strings. It makes some sense for JSON, but for CBOR it's not good. I think we would benefit from picking a default serialization target for `common` structs, and I think it should probably be CBOR. Shout if you think this is the wrong approach

modules/chain_store/src/chain_store.rs
```diff
@@ -733,6 +778,493 @@ impl ChainStore {
...
+                    {
+                        certs.push(TransactionDelegationCertificate {
+                            index: index as u64,
+                            address: map_stake_address(cred, network_id.clone()),
+                            pool_id: to_pool_id(pool_key_hash),
+                            active_epoch: tx.block.extra.epoch + 1,
+                        });
+                    }
+                }
+                _ => (),
```
sandtreader: ... and here...

modules/chain_store/src/chain_store.rs
```diff
@@ -733,6 +778,493 @@ impl ChainStore {
...
+                                cost,
+                                margin,
+                                reward_account,
+                                pool_owners,
+                                relays,
+                                pool_metadata,
+                                network_id.clone(),
+                                false,
+                            )?,
+                            active_epoch: tx.block.extra.epoch + 1,
```
sandtreader: If valid (I assume it is!) needs a comment as to why incremented

modules/rest_blockfrost/src/handlers/transactions.rs
```diff
@@ -0,0 +1,629 @@
...
+            return Err(RESTError::invalid_param(
+                "transaction",
+                "Invalid transaction hash",
+            ))
+        }
+    };
+
+    match param {
+        None => handle_transaction_query(context, tx_hash, handlers_config).await,
+        Some("utxo") => Ok(RESTResponse::with_text(501, "Not implemented")),
```
sandtreader: Is there a ticket to do this?

modules/rest_blockfrost/src/handlers/transactions.rs
```diff
@@ -0,0 +1,629 @@
...
+            TransactionOutputAmount::Asset(asset) => {
+                state.serialize_field("unit", &asset.name)?;
+                state.serialize_field("quantity", &asset.amount.to_string())?;
+            }
+        }
+        state.end()
+    }
+}
+
+/// Handle `/txs/{hash}`
```
sandtreader: Actually /txs/{hash}[/param][/param2]] I think? (to differentiate from the next fn)

modules/rest_blockfrost/src/handlers/transactions.rs
```diff
@@ -0,0 +1,629 @@
...
+                )
+                .await
+                {
+                    Ok(params) => params,
+                    Err(e) => return Some(Err(e)),
+                };
+                // TODO: calc from outputs and inputs if recorded_fee is None
+                let fee = txs_info.recorded_fee.unwrap_or_default();
+                let deposit = match calculate_deposit(
+                    txs_info.pool_update_count,
```
sandtreader: Worried about this - check if deposit is paid for subsequent updates of the same pool reg.

sandtreader (changes requested):
Massive piece of work, good job!

Only queries are about possible missing Conway certs where you can register and delegate at the same time, and whether counting PoolRegistrations is enough for deposits.

modules/rest_blockfrost/src/handlers/transactions.rs
alexwoods: I will add a comment to check for now

PR #373, opened 2025-11-19
feat: address UTxOs of a specific asset endpoint

whankinsiv:
## Description
Adds the REST handler for `/addresses/{address}/utxos/{asset}`. The implementation mirrors the existing `/addresses/{address}/utxos` endpoint, with the addition of a filtering step to return only UTxOs containing the requested asset. 

## Related Issue(s)
Completes #335

## How was this tested?
Compared responses against Blockfrost to confirm:
* all UTxOs containing the specified asset are returned
* an empty array is returned when an address holds none of the specified asset

## Checklist

- [x] CI is green for this PR

## Impact / Side effects
None

The only meaningful changes are in the new endpoint handler.


modules/rest_blockfrost/src/handlers/assets.rs
```diff
@@ -295,28 +296,6 @@ pub async fn handle_policy_assets_blockfrost(
     Ok(RESTResponse::with_json(200, &json))
 }
 
-fn split_policy_and_asset(hex_str: &str) -> Result<(PolicyId, AssetName), RESTError> {
```
lowhung: I think there are a few tests related to this function that should probably be moved into the `utils.rs` as well.

lowhung (approved):

Issue #374, opened 2025-11-19
Make PeerNetworkInterface command-driven


SupernaviX:
To support a custom indexer application, we need to be able to control chainsync programmatically; the indexer should be able to decide where syncing should begin from, at runtime.

- Make PeerNetworkInterface listen for "sync" commands over some new topic. When it receives one of these commands, it should reset the chain to the specified point.
- Make the PeerNetworkInterface `sync-point` configuration setting optional. If it's not specified, the module should not start running chainsync until it receives a "sync" command.

PR #394, opened 2025-11-20
feat: Add dynamic sync mode to peer_network_interface

whankinsiv:
## Description
This PR introduces a new `dynamic` sync mode to the `peer_network_interface`. In this mode, the interface waits for a `ChangeSyncPoint` command on the `cardano.sync.command` topic and begins fetching from that point. 

A background task forwards incoming sync points to the `NetworkManager`. When a new point arrives, the manager resets the `ChainState`, clears outstanding peer requests, and issues a new `find_intersect` across all upstream peers. 

A minimal `indexer` process is included to demonstrate this behavior. It registers only the required modules and emits two `ChangeSyncPoint` commands 5 seconds apart, demonstrating dynamic re-syncing. 

## Related Issue(s)
#374

## How was this tested?
By running the `indexer` process:
1. It publishes a sync point at the last Byron block. 
2. The `peer_network_interface` begins fetching blocks normally. 
3. A second `ChangeSyncPoint` is published, returning to the starting point. 
4. The `peer_network_interface` resets and syncs again from that point. 

For visibility, you can log inside `handle_roll_forward` to observe block fetching. 

## Checklist
- [x] CI is green for this PR

## Impact / Side effects
Adds a basic `indexer` process intended to become the minimal working example for building custom indexers on top of Acropolis. 

* Dynamic `NetworkManager` construction in `peer_network_interface/src/peer_network_interface.rs`. 
* Correct behavior of `on_sync_cmd` when resetting sync state in `peer_network_interface/src/network.rs`.

### DIFF
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -29,7 +29,6 @@ regex = "1"
 serde = { workspace = true, features = ["rc"] }
 serde_json = { workspace = true }
 serde_with = { workspace = true, features = ["base64"] }
-tempfile = "3"
 tokio = { workspace = true }
 tracing = { workspace = true }
 futures = "0.3.31"
@@ -40,8 +39,11 @@ rayon = "1.11.0"
 cryptoxide = "0.5.1"
 thiserror = "2.0.17"
 sha2 = "0.10.8"
+
+[dev-dependencies]
 caryatid_process = { workspace = true }
 config = { workspace = true }
+tempfile = "3"
 
 [lib]
 crate-type = ["rlib"]


modules/indexer/src/indexer.rs
```diff
@@ -0,0 +1,69 @@
...
+use caryatid_sdk::{module, Context, Module};
+use config::Config;
+use std::{str::FromStr, sync::Arc};
+use tracing::info;
+
+// Configuration defaults
+const DEFAULT_DYNAMIC_SYNC_TOPIC: (&str, &str) =
+    ("dynamic-sync-publisher-topic", "cardano.sync.command");
+
+/// Historical Epochs State module
```
SupernaviX:
```suggestion
/// Indexer module
```

modules/peer_network_interface/src/peer_network_interface.rs
```diff
@@ -118,6 +126,36 @@ impl PeerNetworkInterface {
...
+                        Ok(Point::Specific(slot, hash)) => {
+                            let (epoch, _) = sink.genesis_values.slot_to_epoch(slot);
+                            sink.last_epoch = Some(epoch);
+                            info!("Dynamic sync starting at slot {} (epoch {})", slot, epoch);
+                            Point::Specific(slot, hash)
+                        }
+                        Ok(Point::Origin) => {
+                            warn!("Dynamic sync received Point::Origin; ignoring");
+                            return;
+                        }
```
SupernaviX: I don't think there's any reason to not support syncing to origin.

modules/peer_network_interface/src/peer_network_interface.rs
```diff
@@ -44,6 +45,13 @@ impl PeerNetworkInterface {
             _ => None,
         };
 
+        // Create background task to foward sync commands to NetworkManager
+        let mut cmd_rx = if cfg.sync_point == SyncPoint::Dynamic {
+            Some(Self::spawn_command_forwarder(context.clone(), &cfg.sync_command_topic).await?)
+        } else {
+            None
+        };
```
SupernaviX: Having a dedicated channel receiving sync commands works, but I think that these commands should be sent to `NetworkManager` by writing to its event queue instead.

1. Future requirements: there are going to be more commands in the future, the consensus module will be able to tell this module to follow a different chain/peer and I suspect someday a "peer discovery" module will tell it to connect to new peers or disconnect from shady ones. So this module will have to handle several kinds of commands from outside, and that'll be less onerous if we do it by publishing them all to one queue.

2. I don't think we need to restrict "sync commands" so that they only work when you're using this "dynamic" sync point. We can change which point we sync to at runtime, which means that we could _start_ by syncing to the tip and switch to something else later. I don't have a concrete use case in mind for that, but in general I try not to block functionality if it's easy to support.

modules/indexer/src/indexer.rs
SupernaviX: oh lol you got this one while I was typing

whankinsiv: Fixed in c1ae681.

modules/peer_network_interface/src/peer_network_interface.rs
whankinsiv: Refactored the module to follow this approach in d9110c1. Sync commands are now forwarded into the `NetworkManager` through the same event queue used for peer updates.

modules/peer_network_interface/src/peer_network_interface.rs
whankinsiv: Fixed in d9110c1.

Issue #397, opened 2025-11-21
Introduce new field to BlockInfo structure


shd:
Add 'intent' field to BlockInfo structure, as an enum with two variants: 'Validate' and 'Apply'.

This field is to be generated by mithril/upstream fetcher as 'Apply' (important for the ticket).
And in consensus module it should be generated as 'Validate' (may be easily left out of the ticket).

Later (outside the scope of this ticket) this field is to be checked by other modules, and maybe they would have different behaviour based on that (e.g. update internal structures or send validation messages).

Expected outcome: the project works like before with upstream.

It's not important to fully make correct PR, just have something compiling/working in github to allow work on other modules.

alexwoods:
Can be both be false? Because if not, an enum is safer

PR #412, opened 2025-11-21
Add intent field to BlockInfo

alexwoods:
## Description

This PR adds a new field to BlockInfo - intent. This signals whether the block should be just validated, just applied or both.

## Related Issue(s)
Relates to #397.

## How was this tested?
The only test for this change is that it doesn't break anything, so this was tested using existing unit tests and by running the service to make sure it still functions.

## Checklist

- [x] CI is green for this PR

## Impact / Side effects
When creating new blocks the intent field must be set.

Nothing specific.


modules/block_vrf_validator/src/ouroboros/praos.rs
```diff
@@ -113,6 +113,7 @@ mod tests {
             hex::decode(include_str!("./data/7854823.cbor")).unwrap();
         let block_info = BlockInfo {
             status: BlockStatus::Immutable,
+            intent: BlockIntent::Apply,
```
sandtreader: Given this is a validator, shouldn't this be BlockIntent::Validate?

lowhung (approved):

PR #415, opened 2025-11-24
Comments about ledger verification

shd:
Outcome of issue 413: motivation for the newly created tickets.

### DIFF
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -29,6 +29,7 @@ regex = "1"
 serde = { workspace = true, features = ["rc"] }
 serde_json = { workspace = true }
 serde_with = { workspace = true, features = ["base64"] }
+tempfile = "3"
 tokio = { workspace = true }
 tracing = { workspace = true }
 futures = "0.3.31"
@@ -39,6 +40,8 @@ rayon = "1.11.0"
 cryptoxide = "0.5.1"
 thiserror = "2.0.17"
 sha2 = "0.10.8"
+caryatid_process = { workspace = true }
+config = { workspace = true }
 
 [lib]
 crate-type = ["rlib"]


docs/ledger-verification.md
```diff
@@ -0,0 +1,87 @@
...
+1. Blocks and transactions can appear in Acropolis from two sources:
+   * Mithril/other trusted source --- blocks, already accepted by the blockchain.
+     If everything is ok, then the block is applied, internal structures updated, and next block
+     is processed.
+     If something is not correct, then the whole blockchain is broken, and outside intervention
+     is required.
+
+   * Mempool or consensus blocks --- proposals for the blockchain. If the block/transaction is not 
+     successfully verified, then it could be refused.
+     If block is verified, it may be included into chain as mutable (optimistic version),
```
sandtreader: Yes it is consensus' choice whether to send it twice with Validate and Apply (two-phase commit style) or the combined one (optimising single phase commit).  In the latter case if it fails it will send a rollback flag set when the next block candidate is sent out.

docs/ledger-verification.md
```diff
@@ -0,0 +1,87 @@
...
+
+2. The ledger always checks transactions for all errors and performs all necessary checks.
+But the reaction to the errors can be different.
+
+# Two types of reaction
+
+We have basically two variants of behaviour if the error in application/verification occurred:
+
+1. Print an error message. This happens when an incorrect block apply happened, which probably
+   broke the state. We can only hope that further blockchain rollbacks will correct it.
```
sandtreader: Yes, if we treat the content of Mainnet as Gospel, then any errors we get from normal replay of it are our fault.

sandtreader (approved):
Really good description of the plan!  It raises a good point about cross-module inconsistency in the event of a failed validation - at the moment the rollback can only be triggered by sending a new block.

Also that consensus cannot rely on receiving N responses if one of them is a NoGo - others may be dependent on it.  I'll raise a ticket to fix this.

I will fold this doc into the architecture docs when I come to it, but this is a good placeholder in the meantime!

Issue #421, opened 2025-11-25
Add explicit rollback messages through the system


SupernaviX:
Add a specific new Rollback message, written to every topic which is carrying state changes. Most consumers should propagate this message downstream and otherwise ignore it. The rollback status on blockinfo should keep the semantics it has today.

PR #423, opened 2025-11-26
Introduce new rollback strategy

SupernaviX:
## Description

Right now, all the state modules handle rollbacks at the same time that they process new messages. After a rollback has occurred, the `status` field on the next `BlockInfo` will be `RolledBack`. When modules read this status from messages on their "synchronizer" topic (the topic they use to decide where on the chain they are), they roll back all state to right before the newly received message before processing it.

The trouble with this is
 1. If the chain rolls _back_, modules won't react to the rollback until the next time it rolls _forward_. That leaves our node in an inconsistent state until that roll forward.
 2. Nothing in the system signals which block we've rolled back _to_.  This would be a problem for indexers, which need to know how far they've indexed for operational reasons.

### What this PR does

Introduce a new StateTransition message, with a Rollback variant. The system is prepared to send Rollback state transition messages throughout itself. When a module receives a Rollback on its "synchronizer" topic, it will forward that Rollback to every topic it writes to. When a module receives a Rollback on any other topic it reads from, it will ignore it.

I've added two utilities to make this pattern easier to follow:
 - `Subscription.read_ignoring_rollbacks` will read from a subscription until it finds any message which is not a Rollback, and then return that.
 - `RollbackAwareProducer` writes messages to a specific topic, but will only write Rollback messages to that topic if they they would roll back the most recently published non-rollback message. The idea is that we only propagate rollback messages to downstream topics if there is anything from the topic to actually roll back.

I also added a `Point` type to use in the point message, and made `PeerNetworkInterface` accept that type in its `FindIntersect` command. That's because rollbacks can theoretically be all the way to the origin, and indexing could start at the origin.

### What this PR does not do

Nothing is actually producing these new state transition messages yet. The PeerNetworkInterface will eventually publish them to `cardano.block.available`, but it doesn't yet.

Nothing depends on state transitions. They are forwarded around the system, but no functionality depends on them. Everything still relies on checking for `block_info.status == RolledBack`.

## Related Issue(s)

Contributes to #421 

## How was this tested?

Ran omnibus, didn't see any behavioral differences from before. Which is what I expected, because the new messages aren't actually threaded through.

## Checklist

- [x] CI is green for this PR

## Impact / Side effects

None

I think the overall architecture is more relevant in this PR than any specific code.


modules/governance_state/src/governance_state.rs
```diff
@@ -229,7 +218,22 @@ impl GovernanceState {
         });
 
         loop {
-            let (blk_g, gov_procs) = Self::read_governance(&mut governance_s).await?;
+            let (_, message) = governance_s.read().await?;
+            //let (blk_g, gov_procs) = Self::read_governance(&mut governance_s).await?;
```
sandtreader: Commented code alert...  Is read_governance() no longer used?

modules/accounts_state/src/spo_distribution_publisher.rs
```diff
@@ -25,18 +20,19 @@ impl SPODistributionPublisher {
...
+                block.clone(),
+                CardanoMessage::SPOStakeDistribution(SPOStakeDistributionMessage {
+                    epoch: block.epoch - 1, // End of the previous epoch
+                    spos: spos.into_iter().collect(),
+                }),
+            ))))
             .await
     }
+
+    /// Publish a rollback message, if we have anything to roll back
```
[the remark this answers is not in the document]
SupernaviX: There's no formal concept of a "publisher" in this codebase at all, just a bunch of existing structs named Publisher with similar logic. I moved that logic into a `RollbackAwarePublisher` struct, which all of the existing structs are now just wrappers for.

modules/governance_state/src/governance_state.rs
SupernaviX: Removed.

`read_governance` is no longer used, because now this stream can emit two different messages (`CardanoMessage::GovernanceProcedures` or `CardanoMessage::Rollback`) and we want to handle both of them. Hiding the match-on-message-type logic inside a helper function would have made that complicated.

PR #438, opened 2025-12-02
Undelegate stake from retiring pools

### LINKED ISSUE #441: Clear stake delegations to retired pools
When a pool retires, the stake_addresses map in AccountState is not purged of delegations to it.  This leaves apparent active stake which shouldn't be there.

### Expected result

Invariant:  The stake address map should never contain delegations to pools that have retired.

sandtreader:
Should fix divergence in active stake calculations

Also added special cut-down config to only run core functions, for reward and performance testing

$ cargo run --release -- --config omnibus-rewards.toml

## Description

Ensure that when a pool retires, all stake addresses that delegate to it are marked as undelegated.  This will affect
the 'active_stake' totals

## Related Issue(s)
Related to, but does not fix, #8 

## How was this tested?
Additional unit test, needs verification against active stake output from Haskell (@shd has the figures)

## Checklist

- [ X] My code builds and passes local tests
- [ X] I added/updated tests for my changes, where applicable
- [ X] CI is green for this PR

## Impact / Side effects
-
-

Fixes #441

### DIFF
--- a/common/src/stake_addresses.rs
+++ b/common/src/stake_addresses.rs
@@ -448,6 +448,15 @@ impl StakeAddressMap {
         }
     }
 
+    /// Remove all delegations to a given SPO
+    pub fn remove_all_delegations_to(&mut self, spo: &PoolId) {
+        for sas in self.inner.values_mut() {
+            if sas.delegated_spo.as_ref() == Some(spo) {
+                sas.delegated_spo = None;
+            }
+        }
+    }
+
     /// Record a drep delegation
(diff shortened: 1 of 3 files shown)


golddydev (approved):

sandtreader:
> SPDD broadast at epochs 209,210,211 are equal to Haskell node data.

@shd Great!  I take it that was not the case before?  Does it then break at 212?  We think the rewards paid are correct up to 215.

PR #446, opened 2025-12-03
feat:Implement snapshot parser for snapshot delegation and pool registration

### LINKED ISSUE #450: Extend bootstrap parser for snapshot delegations and stake pools
## What
As a developer working on "boot from snapshot", I require a new callback for the Mark/Set/Go snapshot data contained in the stage *delegation* and *pool* sections.

## Acceptance Criteria

- [x] Parser correctly reads the delegation and pool data of each of the Mark/Set/Go snapshot sections
- [x] Parameters are added for the snapshot callback that utilize existing datatypes (no adapters required)
- [x] Sample test driver is extended to prove parsed values
- [x] Bootstrapper callback API is updated with new parameters with a TODO marker for future work

buddhisthead:
## Description

Implement snapshot parser for snapshot delegation and pool registration
* Implemented parser and type conversions for pool registration
* Parsing delegations
* Add logging output for new snapshot data to the test example

## Related Issue(s)
Resolves #450

## How was this tested?
Tested manually with the example/test_streaming_parser executable.
From the top-level, run `make snap-test-streaming` and confirm there are unique values for the Mark Set and Go snapshot output.

## Checklist

- [x] CI is green for this PR

## Impact / Side effects
Describe any potential side effects, e.g. performance, compatibility, or security concerns.
If the PR introduces a breaking change, explain migration steps for users.

If you want specific feedback, list files/functions to review or aspects you are unsure about.

### DIFF
--- a/common/examples/test_streaming_parser.rs
+++ b/common/examples/test_streaming_parser.rs
@@ -275,28 +275,67 @@ impl SnapshotCallbacks for CountingCallbacks {
 impl SnapshotsCallback for CountingCallbacks {
     fn on_snapshots(&mut self, snapshots: RawSnapshotsContainer) -> Result<()> {
         eprintln!("Raw Snapshots Data:");
+        eprintln!();
 
         // Calculate total stakes and delegator counts from VMap data
         let mark_total: i64 = snapshots.mark.0.iter().map(|(_, amount)| amount).sum();
         let set_total: i64 = snapshots.set.0.iter().map(|(_, amount)| amount).sum();
         let go_total: i64 = snapshots.go.0.iter().map(|(_, amount)| amount).sum();
 
-        eprintln!(
-            "  Mark snapshot: {} delegators, {} total stake (ADA)",
-            snapshots.mark.0.len(),
-            mark_total as f64 / 1_000_000.0
-        );
-        eprintln!(
-            "  Set snapshot: {} delegators, {} total stake (ADA)",
-            snapshots.set.0.len(),
-            set_total as f64 / 1_000_000.0
-        );
-        eprintln!(
-            "  Go snapshot: {} delegators, {} total stake (ADA)",
-            snapshots.go.0.len(),
-            go_total as f64 / 1_000_000.0
-        );
-        eprintln!("  Fee: {} ADA", snapshots.fee as f64 / 1_000_000.0);
+        eprintln!("Mark Snapshot:");
+        eprintln!("  Delegators: {}", snapshots.mark.0.len());
+        eprintln!("  Total stake: {:.2} ADA", mark_total as f64 / 1_000_000.0);
+        if !snapshots.mark.0.is_empty() {
+            eprintln!("  Sample stakes (first 5):");
+            for (i, (cred, amount)) in snapshots.mark.0.iter().take(5).enumerate() {
+                let cred_str = cred.to_string().unwrap_or_default();
+                eprintln!(
+                    "    [{}] {} -> {:.2} ADA",
+                    i + 1,
+                    cred_str,
+                    *amount as f64 / 1_000_000.0
+                );
+            }
+        }
+        eprintln!();
+
+        eprintln!("Set Snapshot:");
+        eprintln!("  Delegators: {}", snapshots.set.0.len());
+        eprintln!("  Total stake: {:.2} ADA", set_total as f64 / 1_000_000.0);
+        if !snapshots.set.0.is_empty() {
+            eprintln!("  Sample stakes (first 5):");
+            for (i, (cred, amount)) in snapshots.set.0.iter().take(5).enumerate() {
+                let cred_str = cred.to_string().unwrap_or_default();
+                eprintln!(
+                    "    [{}] {} -> {:.2} ADA",
+                    i + 1,
+                    cred_str,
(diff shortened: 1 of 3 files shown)


common/examples/test_streaming_parser.rs
```diff
@@ -290,28 +275,88 @@ impl SnapshotCallbacks for CountingCallbacks {
...
-            go_total as f64 / 1_000_000.0
-        );
-        eprintln!("  Fee: {} ADA", snapshots.fee as f64 / 1_000_000.0);
+        eprintln!("Mark Snapshot:");
+        eprintln!("  Delegators: {}", snapshots.mark.0.len());
+        eprintln!("  Total stake: {:.2} ADA", mark_total as f64 / 1_000_000.0);
+        if !snapshots.mark.0.is_empty() {
+            eprintln!("  Sample stakes (first 5):");
+            for (i, (cred, amount)) in snapshots.mark.0.iter().take(5).enumerate() {
+                let cred_str = match cred {
```
lowhung: Nit: I know this is just for testing, but the `StakeCredential` type has a `to_string()` implementation that you could use instead of doing this matching on the type of credential in a few places.

```rust
impl StakeCredential {
    pub fn to_string(&self) -> Result<String> {
        let (hrp, data) = match &self {
            Self::AddrKeyHash(data) => (Hrp::parse("stake_vkh")?, data.as_slice()),
            Self::ScriptHash(data) => (Hrp::parse("script")?, data.as_slice()),
        };

        Ok(bech32::encode::<Bech32>(hrp, data)?)
    }
}
```

I think the whole bech-32 encoding would be more useful for debugging than using 8 hex characters in this case.

common/src/snapshot/mark_set_go.rs
```diff
@@ -140,104 +110,166 @@ impl Snapshot {
...
+                            .context("Failed to parse snapshot_pool_registration")?;
+                        let registration = VMap(
+                            pools
+                                .0
+                                .into_iter()
+                                .map(|(pool_id, params)| {
+                                    // Convert RewardAccount (Vec<u8>) to StakeAddress
+                                    let reward_account =
+                                        StakeAddress::from_binary(&params.reward_account.0)
+                                            .unwrap_or_else(|_| StakeAddress::default());
```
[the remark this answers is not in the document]
lowhung: Unless the snapshot is corrupted, I wouldn't be too concerned about this `from_binary` failing, but if it did fail for some reason...we would definitely want to know about it.

lowhung (commented):
Nice work @buddhisthead! Main concern is the `AddrKeyHash` parse.

common/src/snapshot/mark_set_go.rs
```diff
@@ -77,11 +82,10 @@ pub trait SnapshotsCallback {
 pub struct Snapshot {
     /// snapshot_stake: stake distribution map (credential -> lovelace amount)
     pub snapshot_stake: VMap<StakeCredential, i64>,
-    // snapshot_delegations: delegation map (credential -> stake pool key hash)
-    // pub snapshot_delegations: VMap<Credential, StakePool>,
-
-    // snapshot_pool_params: pool parameters map (stake pool key hash -> pool params)
-    //  snapshot_pool_params: VMap<Vec<u8>, Vec<u8>>,
+    // snapshot_delegations: vmap<credential, key_hash<stake_pool>>,)
+    pub snapshot_delegations: VMap<StakeCredential, Hash<28>>,
```
lowhung: I think this could use `PoolId` instead, given it's a map of staking credential to the pool to which they're delegated to.


```suggestion
    pub snapshot_delegations: VMap<StakeCredential, acropolis_common::PoolId>,
```

common/src/snapshot/mark_set_go.rs
```diff
@@ -77,11 +82,10 @@ pub trait SnapshotsCallback {
...
+    // snapshot_pool_params: vmap<key_hash<stake_pool>, pool_params>,
+    pub snapshot_pool_params: VMap<Hash<28>, PoolRegistration>,
```
lowhung: I think this can also be 
```suggestion
    pub snapshot_pool_params: VMap<PoolId, PoolRegistration>,
```

common/examples/test_streaming_parser.rs
buddhisthead: Agreed. I missed the to_string() implementation, but I've changed this to use it - nice cleanup! Thanks.

common/src/snapshot/mark_set_go.rs
```diff
@@ -140,104 +110,166 @@ impl Snapshot {
...
+                                            .unwrap_or_else(|_| StakeAddress::default());
+
+                                    // Convert Set<AddrKeyhash> to Vec<StakeAddress>
+                                    let pool_owners: Vec<StakeAddress> = params
+                                        .owners
+                                        .0
+                                        .into_iter()
+                                        .map(|keyhash| {
+                                            StakeAddress::new(
+                                                StakeCredential::AddrKeyHash(keyhash),
```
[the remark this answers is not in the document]
buddhisthead: IN a discussion with @lowhung we decided there wasn't a way to determine a choice of encoding, so we're just documenting the choice in the code.

Issue #410, opened 2025-11-21
Split TxUnpacker and use UTxO status module


shd:
1. Remove all use of UTxO storage from TxUnpacker (generate raw transaction types, which can be constructed without need resolve UTxO references)
2. Move all UTxO resolving to UTxO module.
3. Make UTxO storage module posting to transactions channels.
4. Make TxValidatorPhase1 receiving UTxO transaction outcomes.

golddydev:
@shd This issue is done by https://github.com/input-output-hk/acropolis/pull/451 ?

PR #451, opened 2025-12-04
Switch UTXOIdentifier back to being (tx_hash, output_index)

### LINKED ISSUE #448: Revert UTxOIdentifier to use TxHash
## Context

The `tx_unpacker` maintains a `UTxORegistry` that tracks the live UTxO set, mapping `TxOutRef → TxIdentifier`. This enables converting transaction inputs into `UTxOIdentifier` (block_number + tx_index + output_index) for downstream modules like `utxo_state`.

When processing a transaction, the `tx_unpacker`:
1. Looks up inputs in its registry to get the `TxIdentifier` (to create `UTxOIdentifier`)
2. Removes those inputs from the registry (they're spent)
3. Adds the new outputs to the registry

## Problem

When bootstrapping from a snapshot, we bypass `tx_unpacker` entirely:
- `utxo_state` gets the full UTxO set from the snapshot ✓
- `tx_unpacker`'s registry remains empty ✗

When new blocks arrive post-snapshot, `tx_unpacker` can't resolve inputs to `UTxOIdentifier` because they never passed through it. The `tx_unpacker` doesn't crash—it silently drops unresolvable inputs and logs an error. The result is incomplete delta messages where spent UTxOs never get marked as spent in `utxo_state`.

We cannot bootstrap the `UTxORegistry` from the snapshot because it doesn't contain `block_number` or `tx_index`—only `TxOutRef` (tx_hash + output_index).

## Options Discussed

1. Revert to TxHash-based identifiers — Simpler and safer; removes dependency on block_number/tx_index. Need to assess memory impact on other modules (e.g., Address State)

2. Invent synthetic serial numbers — Pass snapshot through `tx_unpacker` with invented identifiers. Unclear if other components construct `UTxOIdentifier` from block_number independently.

3. Bootstrap the registry somehow — Not feasible without the required data in snapshots.

## Additional Consideration

The `/tx/utxos` endpoint requires data not immediately available in the requested transaction:
- Input UTxO values (chain_store could look these up)
- Hash of tx that consumed outputs (currently falls through the gaps)

## Acceptance Criteria

- [ ] Snapshot bootstrap populates `utxo_state` directly without breaking input resolution
- [ ] Post-snapshot blocks correctly mark spent UTxOs
- [ ] Determine memory impact of TxHash-based approach on affected modules
- [ ] Clarify ownership of consumed_by_tx lookup between chain_store and utxo_state

sandtreader:
Reverts UTXOIdentifier to being (tx_hash, output_index), enabling bootstrap from snapshot.

Work In Progress!

## Description

Reverts UTXOIdentifier - and hence storage key in UTXOState - to being tx hash based, removing memory optimisation through UTXORegistry in TxUnpacker (which has been removed).

This also means the REST queries need to look up TX hashes to get block hashes and tx indexes within the block - there is a new GetBlockHashesAndIndexOfTransactionHashes query to do this, but it needs implementing.  I'm not sure if ChainStore has the indexes to do this.

## Related Issue(s)
Fixes #448 #410 

## How was this tested?

Not yet tested - multiple test builds broken at the moment

## Checklist

- [x] CI is green for this PR

## Impact / Side effects

The memory use of AddressState will explode with this change - requires either spilling to disk and/or using CompactUTxOIdentifier internally and then a restituted GetUTXOHashes to look up as before.

### DIFF
--- a/modules/chain_store/src/chain_store.rs
+++ b/modules/chain_store/src/chain_store.rs
@@ -18,10 +18,10 @@ use acropolis_common::{
     },
     queries::{
         blocks::{
-            BlockHashes, BlockInfo, BlockInvolvedAddress, BlockInvolvedAddresses, BlockKey,
-            BlockTransaction, BlockTransactions, BlockTransactionsCBOR, BlocksStateQuery,
-            BlocksStateQueryResponse, NextBlocks, PreviousBlocks, TransactionHashes,
-            DEFAULT_BLOCKS_QUERY_TOPIC,
+            BlockHashAndTxIndex, BlockHashes, BlockInfo, BlockInvolvedAddress,
+            BlockInvolvedAddresses, BlockKey, BlockTransaction, BlockTransactions,
+            BlockTransactionsCBOR, BlocksStateQuery, BlocksStateQueryResponse, NextBlocks,
+            PreviousBlocks, TransactionHashes, DEFAULT_BLOCKS_QUERY_TOPIC,
         },
         misc::Order,
     },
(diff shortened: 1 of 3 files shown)


sandtreader:
Implementing `BlocksStateQuery::GetBlockHashesAndIndexOfTransactionHashes` in ChainStore should fix Clippy errors and get `/address` endpoints working again. @alexwoods

Issue #422, opened 2025-11-25
Add a tip status


SupernaviX:
Add a tip status to blockinfo, for blocks which are at the tip of the chain.

PR #454, opened 2025-12-04
Report tip status in BlockInfo

SupernaviX:
## Description

Add a `tip_slot` field to `BlockInfo`, indicating which slot "the tip" is currently at. This field is `None` by default and while e.g. syncing from mithril, but we populate it while transmitting blocks from the ChainSync mini-protocol. This field currently refers to the tip of the chain of whichever peer we've decided to follow.

Also adds a test for a regression from an older implementation of PeerNetworkInterface.

## Related Issue(s)

Completes #422 

## How was this tested?

I synced the node with preview after booting from a mithril snapshot, confirming that
 - it did not report a tip while restoring from the snapshot
 - it reported a tip in the future while catching up
 - it reported a tip in the present (or past) once caught up

## Checklist

- [x] CI is green for this PR

## Impact / Side effects

An extra optional u64 in BlockInfo which gets passed everywhere

The only non-mechanical changes are updates to BlockInfo and PeerNetworkInterface

### DIFF
--- a/modules/peer_network_interface/src/chain_state.rs
+++ b/modules/peer_network_interface/src/chain_state.rs
@@ -899,7 +899,7 @@ mod tests {
         let (h2, b2) = make_block(11, "second block");
         let (h3, b3) = make_block(12, "third block");
 
-        // Meanwhile, another upstream is sending us blocks. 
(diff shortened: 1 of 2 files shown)


whankinsiv (approved):
Looks good!

Issue #380, opened 2025-11-19
Support multiple indexes in CustomIndexer module


SupernaviX:
Blocked by #375 

Different indexes might be at different points on-chain.

Make sure that indexes receive the correct events when more than one index is defined (i.e. if index A is at height 1000 and index B is at height 1500, index A should be notified about TXs from block 1001 and index B should not.)

Make sure that if one index returns an error while indexing, other indexes continue to advance.

PR #455, opened 2025-12-05
feat: custom indexer with multi index support

whankinsiv:
## Description
This PR extends the `custom_indexer` module to support multiple independent indexes within a single indexer instance. Each index is registered using `add_index`, which creates a dedicated channel for receiving transaction and rollback events. All messages include a oneshot response channel, allowing the manager task to update cursor state and remove corrupted indexes.

The error handling model is explicit and prevents corrupted index state from persisting across runs:
- Decode failure: Indicates invalid transaction bytes (should never occur for valid blocks). The index is halted immediately and awaits the next rollback event. 
- Handle failure: Indicates an error in user-defined index logic. The index is halted and will be reset on the next process start due to the persisted `halted = true` flag. 
- Rollback failure: If an index cannot roll back cleanly, the index attempts a full reset.
  - If reset succeeds, the chain sync point is rewound. 
  - If reset fails, the index is removed from the runtime (`senders`). On next startup, the persisted halt flag causes a fresh reset. 

## Related Issue(s)
Completes #380 

## How was this tested?
* Added full coverage for `index_actor` (apply tx, rollback, reset paths). 
* Verified that the example indexer runs multiple indexes with different starting point correctly. 

## Checklist

- [x] CI is green for this PR

## Impact / Side effects
Adds safe recovery behavior and multi-index support to the custom indexer.

Rollback failure behavior and the halt/persist/reset flow

### DIFF
--- a/modules/custom_indexer/src/custom_indexer.rs
+++ b/modules/custom_indexer/src/custom_indexer.rs
@@ -36,6 +36,10 @@ use crate::{
 
 type IndexSenders = HashMap<String, mpsc::Sender<IndexCommand>>;
 type SharedSenders = Arc<Mutex<IndexSenders>>;
+type IndexResponse = (
+    String,
+    Result<IndexResult, tokio::sync::oneshot::error::RecvError>,
+);
 
 struct IndexWrapper {
     index: Box<dyn ChainIndex>,
@@ -81,7 +85,7 @@ impl<CS: CursorStore> CustomIndexer<CS> {
         if force_restart || entry.halted {
             index.reset(&default_start).await?;
             entry.tip = default_start.clone();
-            entry.halted = false;
+            entry.halted = true;
         }
 
         cursors.insert(name.clone(), entry.clone());
@@ -173,10 +177,10 @@ where
                                     .collect();
 
                                 // Send txs to all index tasks
-                                let mut responses = send_txs_to_indexers(&senders, block, &txs).await;
+                                let responses = send_txs_to_indexers(&senders, block, &txs).await;
 
                                 // Get responses with new tips and any halts that occured
-                                let new_entries = process_tx_responses(&mut responses, block.slot).await;
+                                let new_entries = process_tx_responses(responses, block.slot).await;
 
                                 // Save the new entries to the cursor store
                                 cursor_store.save(&new_entries).await?;
@@ -215,7 +219,7 @@ where
                                     }
                                 }
                             }
-                            _ => (),
+                            _ => error!("Unexpected message type: {message:?}"),
                         }
                     }
                     Err(e) => {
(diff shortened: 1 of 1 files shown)


modules/custom_indexer/src/cursor_store.rs
```diff
@@ -1,79 +1,154 @@
...
+        Ok(Self { partition })
     }
 }
 
 impl CursorStore for FjallCursorStore {
-    async fn load(&self) -> Result<Option<Point>> {
-        let raw = self.cursor.get("cursor")?;
+    async fn load(&self) -> Result<HashMap<String, CursorEntry>> {
+        let mut out = HashMap::new();
+        let iter = self.partition.prefix("cursor/");
```
lowhung: [nit] We reference this prefix `"cursor/"` in a few places, should we keep it as a const in this file?

modules/custom_indexer/src/cursor_store.rs
```diff
@@ -1,79 +1,154 @@
...
     }
 
-    async fn save(&self, point: &Point) -> Result<()> {
-        let raw = bincode::serialize(point)?;
+    async fn save(&self, tips: &HashMap<String, CursorEntry>) -> Result<(), CursorSaveError> {
+        let mut failed = Vec::new();
 
-        self.cursor.insert("cursor", raw)?;
+        for (name, point) in tips {
+            let key = format!("cursor/{name}");
```
lowhung: [nit] A method in the cursor store to format the key and return it / get the prefix could be nice.

modules/custom_indexer/src/custom_indexer.rs
```diff
@@ -7,47 +7,134 @@
...
+        }
+
+        let mut cursors = self.cursor_store.load().await?;
+
+        let mut entry = cursors.get(&name).cloned().unwrap_or(CursorEntry {
+            tip: default_start.clone(),
+            halted: false,
+        });
+
+        if force_restart || entry.halted {
```
lowhung: Nice 👍🏻

lowhung (approved):
Just so I understand the flow...

1. When a new index is added, it spawns an actor per index, each with its own channel
2. ChainSync messages come in on the subscription and fan out to all actors
3. The actors process independently, then return a result with updated cursor entries
4. Main loop collects responses + persists cursors + handles failures

modules/custom_indexer/src/cursor_store.rs
```diff
@@ -1,79 +1,154 @@
...
+        if failed.is_empty() {
+            Ok(())
+        } else {
+            Err(CursorSaveError { failed })
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct CursorSaveError {
```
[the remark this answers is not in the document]
whankinsiv: Thanks for this suggestion! Switched to using `thiserror` in 7b5d429.

PR #480, opened 2025-12-11
feat: fix rewards calculations with stake address map in bootstrap

### LINKED ISSUE #388: Bootstrap Accounts Module
### Context
The Accounts State module keeps track of stake addresses, SPO delegations, DRep delegations, rewards, and pot balances. It also maintains three-epoch snapshot history (mark, set, go) for rewards calculations (IIUC) and processes epoch activity to calculate and distribute rewards. It currently expects many message types starting from genesis, building up delegation history and taking snapshots at epoch boundaries.

### Problem
When bootstrapping from a snapshot (e.g., epoch 500), the accounts state module needs to initialize from messages containing parsed snapshot data:
- Stake addresses with rewards and delegations (SPO and DRep)
- SPO registrations with parameters and retirement info
- DRep registrations with deposits
- Pot balances (treasury, reserves, deposits)
- Three-epoch snapshot history (mark, set, go) for rewards calculation

[The CDDL](https://github.com/rrruko/nes-cddl-hs/blob/main/nes.cddl#L129) we have for the NewEpochState might be useful here.

The bootstrapper will parse the snapshot CBOR, and send the appropriate messages to initialize accounts state. This latter part must be implemented as part of this work.

### Acceptance Criteria
- [x] Bootstrapper parses snapshot and sends message with information required by accounts module
- [x] Accounts module has message handler able to work with message from bootstrapper and initialize state properly
- [ ] Normal epoch processing continues after bootstrap (i.e we can use Mithril or Upstream fetcher post bootstrap)

### Testing Requirements
- [x] Bootstrap data validates against Blockfrost API
- [ ] Come up with other requirements if you can! 
- [ ] Rewards calculations must be correct

### Definition of Done
- Accounts state initializes from snapshot without replaying the entire chain's history
- All stake addresses, delegations, and rewards accurate after bootstrap
- SPO and DRep state accurate after bootstrap
- Pot balances match snapshot values
- Snapshot history properly initialized for rewards calculation
- Blockfrost API validation confirms correctness

lowhung:
## Description

Fixes incorrect reward calculation during snapshot bootstrap. Previously, the bootstrap process used instantaneous reward balances from `DState`, which didn't match the actual stake distribution used by Cardano for epoch calculations. This PR updates the parser to extract rewards from the `pulsing_rew_update` snapshot, ensuring accurate stake distribution and correct SPDD generation.

## Related Issue(s)
Fixes #388 

## How was this tested?
- Ensure `make run` for local snapshot bootstrap with epoch 509 works
- SPDD output verified against cardano-db-sync `epoch_stake` table using (spdd generation code was unstaged, only used as means to test stake address map creation was valid)

```
acropolis/tests/integration on  lowhung/generate-spdd-from-bootstrap [$!?] via  v23.8.0
➜ npm run test:spdd

> acropolis-integration-tests@1.0.0 test:spdd
> ts-node spdd.test.ts

(node:88377) ExperimentalWarning: Type Stripping is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
Validating Acropolis SPDD vs DB Sync...

Epoch 509: ✅ 3023 match, ⚠️ 0 mismatch, ❌ 0 missing, 🌀 0 extra (total 3023)
Reached end of available epochs (HTTP 404).

Finished.
```

## Checklist

- [x] CI is green for this PR

## Impact / Side effects

**Positive**: SPDD now matches cardano-db-sync data, enabling accurate stake calculations and proper node initialization when booting from snapshot. It provides the basis for future utxo / address deltas being applied to the correct base state for the stake address map (which is what the spdd is generated from).

**Performance**: Minimal impact - snapshot parsing is still quick.

- `parse_pulsing_reward_update()` method - handles both pulsing and completed reward states
- Bootstrap message structure changes - ensure downstream consumers handle new fields correctly

### DIFF
--- a/common/src/snapshot/reward_snapshot.rs
+++ b/common/src/snapshot/reward_snapshot.rs
@@ -89,11 +89,9 @@ fn decode_likelihood_value(d: &mut Decoder) -> Result<RationalNumber, minicbor::
         }
     }
 
-    // Try as float or integer - convert to rational representation
     match d.datatype()? {
         Type::F16 | Type::F32 | Type::F64 => {
             // For floats, we approximate as a rational
-            // This is a fallback - ideally all values should be tag 30 rationals
             let f = d.f64()?;
             // Use a large denominator to preserve precision
             let scale = 1_000_000_000_000u64;


common/src/snapshot/utxo.rs
```diff
@@ -0,0 +1,577 @@
...
+    /// - Pointer addresses (stake part is a pointer, not a credential)
+    /// - Reward addresses
+    /// - Invalid/malformed addresses
+    ///
+    /// Address format (Shelley):
+    /// - Header byte encodes network and address type
+    /// - Base addresses (type 0-3): 1 byte header + 28 bytes payment + 28 bytes stake
+    /// - For types 0,1: stake part is key hash
+    /// - For types 2,3: stake part is script hash
+    pub fn extract_stake_credential(&self) -> Option<StakeCredential> {
```
sandtreader: We already have this code in address.rs I think

common/src/snapshot/utxo.rs
```diff
@@ -0,0 +1,577 @@
...
+use crate::StakeCredential;
+
+// =============================================================================
+// TransactionInput
+// =============================================================================
+
+/// Transaction input reference (TxIn)
+/// CDDL: transaction_input = [txin_transaction_id : transaction_id, txin_index : uint .size 2]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct TransactionInput {
```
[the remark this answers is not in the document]
lowhung: Took Alex's work into the `utxo.rs` file instead :)

common/src/snapshot/utxo.rs
lowhung: Good point, I wonder if this would suffice? d8a7a4ae6bac45c829ed4538e3f1726fc715db8e

PR #482, opened 2025-12-12
Remove unused packages to free up disk space for running the tests

jon5000:
## Description

Previous runs of this workflow have failed due to running out of disk space, so we are removing packages which are installed to the base runner image but are not currently required for this project.

The name of the workflow has been updated to indicate that it also runs on PR creation.

## How was this tested?

Tested extensively in fork of repo.
In all tests, github runners had a 72G filesystem, with 18G free following the checkout stage.  Removing these packages gave us 24G free, with 5.6G free after the tests had run, giving us plenty of headroom.

## Impact / Side effects

Some of these packages could potentially be required in the future.  In that case we can either look at different packages to remove, or there are several GB of other installed programs in `/opt` which we could potentially clear.

### DIFF
--- a/.github/workflows/run-tests-on-push-to-main.yml
+++ b/.github/workflows/run-tests-on-push-to-main.yml
@@ -1,4 +1,4 @@
-name: Run tests on push to main
+name: Run tests on push/PR to main
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


hegaleon (approved):
Looks good v.v housekeeping

PR #511, opened 2025-12-17
fix: pulsing rewards, deposits, utxo underflows and pot deltas

lowhung:
## Description

This PR fixes the epoch boundary transition logic when bootstrapping from a snapshot. Previously, the pot values (treasury, reserves, deposits) from the snapshot were being used directly without accounting for the deltas that should be applied at epoch boundary transitions.

Key changes:
- **Pot deltas tracking**: Extract and track `delta_treasury`, `delta_reserves`, and `delta_deposits` from `pulsing_rew_update` and `instantaneous_rewards` in the snapshot CBOR
- **First epoch skip logic**: Skip the normal monetary change calculation on the first epoch after bootstrap, since pot deltas are already extracted from the snapshot
- **Governance deposit handling**: Properly subtract pending and enacted governance proposal deposits from `us_deposited`, as these will be refunded at epoch boundary
- **DRep deposit handling**: Subtract DRep deposits from the total deposits pot
- **Unregistered stake addresses**: Add support for stake addresses that have UTXOs but aren't registered in DState
- **Fee tracking**: Propagate the fee from the snapshot through to the epoch bootstrap message
- **Snapshot assignment fix**: Pots are now assigned to the `set` snapshot (epoch-1) instead of `go` snapshot, matching the ledger semantics

## Related Issue(s)

Relates to #522 

## How was this tested?
- Manual testing with snapshot bootstrap
- SPDD validation test against db-sync (modified to use `live` stake instead of `active`)
- Pot balance verification at epoch boundaries

## Checklist

- [x] CI is green for this PR

## Impact / Side effects
- **Breaking change to `AccountsBootstrapMessage`**: Adds new `pot_deltas` field
- **Breaking change to `SnapshotsContainer`**: Adds new `fee` field
- Some validators are temporarily disabled in `omnibus.toml` (SPO validation commented out)

1. **`common/src/snapshot/streaming_snapshot.rs`**: The core parsing logic for extracting pot deltas from `pulsing_rew_update` and `instant_rewards`. The signed arithmetic for delta application deserves careful review.
2. **`modules/accounts_state/src/state.rs` (lines 384-450)**: The new `skip_first_epoch_rewards` logic and bootstrap pot delta application in `enter_epoch`.
3. **Governance deposit calculation**: The logic at lines 1146-1173 in `streaming_snapshot.rs` that calculates `governance_deposits` by summing pending and enacted proposal deposits.
4. **The hardcoded `GOV_ACTION_DEPOSIT` constant** (100,000 ADA at line 1397) should ideally come from protocol parameters.
5. **Test file changes**: The SPDD test now compares `live` stake instead of `active` - verify this is the intended behavior.

### DIFF
--- a/modules/accounts_state/src/state.rs
+++ b/modules/accounts_state/src/state.rs
@@ -174,38 +174,9 @@ impl State {
         );
 
         // Apply deltas with overflow checks
-        let new_treasury = self.pots.treasury as i128 + deltas.delta_treasury as i128;
-        let new_reserves = self.pots.reserves as i128 + deltas.delta_reserves as i128;
-        let new_deposits = self.pots.deposits as i128 + deltas.delta_deposits as i128;
-
-        if new_treasury < 0 || new_treasury > u64::MAX as i128 {
-            anyhow::bail!(
-                "treasury pot overflow: {} + {} = {}",
-                self.pots.treasury,
-                deltas.delta_treasury,
-                new_treasury
-            );
-        }
-        if new_reserves < 0 || new_reserves > u64::MAX as i128 {
-            anyhow::bail!(
-                "reserves pot overflow: {} + {} = {}",
-                self.pots.reserves,
-                deltas.delta_reserves,
-                new_reserves
-            );
-        }
-        if new_deposits < 0 || new_deposits > u64::MAX as i128 {
-            anyhow::bail!(
-                "deposits pot overflow: {} + {} = {}",
-                self.pots.deposits,
-                deltas.delta_deposits,
-                new_deposits
-            );
-        }
-
-        self.pots.treasury = new_treasury as u64;
-        self.pots.reserves = new_reserves as u64;
-        self.pots.deposits = new_deposits as u64;
+        update_value_with_delta(&mut self.pots.treasury, deltas.delta_treasury)?;
+        update_value_with_delta(&mut self.pots.reserves, deltas.delta_reserves)?;
+        update_value_with_delta(&mut self.pots.deposits, deltas.delta_deposits)?;
 
         info!(
             "Accounts state bootstrap complete for epoch {}: {} accounts, {} pools, {} DReps, \


common/src/snapshot/streaming_snapshot.rs
```diff
@@ -1186,19 +1263,37 @@ impl StreamingSnapshotParser {
         let drep_deposits: Vec<(DRepCredential, u64)> =
             dreps.iter().map(|(cred, record)| (cred.clone(), record.deposit)).collect();
 
+        // Calculate total DRep deposits
+        let total_drep_deposits: u64 = drep_deposits.iter().map(|(_, d)| d).sum();
+        let total_pool_deposits: u64 = (pool_registrations.len() as u64) * 500_000_000; // 500 ADA per pool
+
```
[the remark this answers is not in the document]
whankinsiv: I agree with Copilot here.

common/src/snapshot/streaming_snapshot.rs
```diff
@@ -1128,22 +1199,28 @@ impl StreamingSnapshotParser {
...
 
-        // Build pots for snapshot conversion
+        // Build pots for snapshot conversion (these are the epoch N pots from the snapshot)
         let pots = Pots {
             reserves,
             treasury,
             deposits,
         };
 
+        // Log the deltas from pulsing_rew_update (these will be applied at epoch boundary)
```
whankinsiv: This comment is somewhat misleading. We are applying these to the reward balance in the bootstrapper to replace needing to apply them at the epoch boundary.

common/src/snapshot/streaming_snapshot.rs
```diff
@@ -1225,6 +1348,88 @@ impl StreamingSnapshotParser {
...
+        // Each enacted or expired governance action gets its deposit (100,000 ADA) refunded
+        // TODO: Get actual gov_action_deposit from protocol parameters
+        const GOV_ACTION_DEPOSIT: u64 = 100_000_000_000; // 100,000 ADA
+        let total_gov_proposals_refunded = enacted_proposal_count + expired_proposal_count;
+        let gov_deposit_refunds = (total_gov_proposals_refunded as u64) * GOV_ACTION_DEPOSIT;
+
+        if total_gov_proposals_refunded > 0 {
+            info!(
+                "Governance deposit refunds: {} enacted + {} expired = {} proposals, {} ADA refunded",
+                enacted_proposal_count,
```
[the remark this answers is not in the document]
whankinsiv: This is wrong. We should use the sum of the `proposal_procedure_deposit` fields for enacted/expired actions.

common/src/messages.rs
```diff
@@ -440,15 +440,30 @@ pub struct AccountsBootstrapMessage {
     /// All registered DReps with their deposits (credential, deposit amount)
     pub dreps: Vec<(DRepCredential, u64)>,
 
-    /// Pot balances (treasury, reserves, deposits)
+    /// Pot balances (treasury, reserves, deposits) for the snapshot epoch
```
whankinsiv: *set epoch

modules/accounts_state/src/state.rs
```diff
@@ -362,31 +384,77 @@ impl State {
...
         self.epoch_snapshots.push(snapshot);
 
         // Pay the refunds after snapshot, so they don't appear in active_stake
         reward_deltas.extend(self.pay_pool_refunds());
 
-        // Verify pots state
-        verifier.verify_pots(epoch, &self.pots);
+        // On first epoch after bootstrap, apply the pot deltas from pulsing_rew_update
+        // and instantaneous_rewards before verification
+        let monetary_change = if self.skip_first_epoch_rewards {
```
whankinsiv: I would prefer passing in the skip condition from the main run loop instead of it being a part of state.

whankinsiv (approved):
Stack overflow issue is fixed in 80a197c

PR #512, opened 2025-12-17
Shd/accounts state validation outcomes

### LINKED ISSUE #398: Modify accounts_state to report errors
This ticket should allow ledger to perform block verification.

1. Collect all validation errors into ValidationError struct values, and have a list of them at the end of each message processing.

2. Check Intention enum in BlockInfo:
   a) If it is 'Apply', then work as usual, and errors are to be printed.

   b) If it is 'Verify', then no internal structures are to be updated, and the errors are to be sent to validation channel ('cardano.validation.account'), using ValidationStatus enum. For simplicity, only the first error is to be passed through the channel, although the implementation should collect all known errors, and have them in vector at the send moment.

   If no errors found, ValidationStatus::Go message is mandatory, so module must send always exactly one ValidationStatus message in 'Verify' mode.

   c) if both flags are set, then 'Apply' should take place if 'Verify' was successful (no errors) -- this may be implemented at later stage, if it is a complicated change; then a new ticket should be created, and info about that should be mentioned in the PR/this ticket.

All other functionality (like passing messages further) should work as usual.

shd:
## Description

Adds account_state verification outcomes (instead of errors)

## Related Issue(s)
Fixes #398 

## How was this tested?
Modified unit tests, added integration test (block.0003.huge-withdrawal.cbor block, used by fake_block_injector)

## Checklist

- [ ] My code builds and passes local tests
- [ ] CI is green for this PR

## Impact / Side effects
Changes error reporting and error reaction in accounts_state.

### DIFF
--- a/common/src/stake_addresses.rs
+++ b/common/src/stake_addresses.rs
@@ -615,12 +615,14 @@ mod tests {
             let stake_address = create_stake_address(STAKE_KEY_HASH);
 
             // Create an entry but don't register
-            stake_addresses.process_stake_delta(&StakeAddressDelta {
-                stake_address: stake_address.clone(),
-                addresses: Vec::new(),
-                tx_count: 1,
-                delta: 100,
-            });
+            stake_addresses
+                .process_stake_delta(&StakeAddressDelta {
+                    stake_address: stake_address.clone(),
+                    addresses: Vec::new(),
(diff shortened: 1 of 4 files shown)


sandtreader (approved):
Looks great, just some commented test code to chop out.

Issue #387, opened 2025-11-20
Bootstrap Parameters Module


lowhung:
### Context
The Parameters State module tracks protocol parameters across eras (Byron through Conway) by processing governance outcomes at epoch boundaries. Currently expects `GovernanceOutcome` messages starting from genesis to build up parameter history through era transitions.

### Problem
When bootstrapping from a snapshot (e.g., epoch 500), the parameters state module needs to initialize with the correct protocol parameters for that epoch rather than replaying all governance outcomes from genesis. According to [the CDDL specification](https://github.com/rrruko/nes-cddl-hs/blob/main/nes.cddl#L129), protocol parameters are stored in the snapshot at:
- Current parameters: `gs_current_pparams` at `NewEpochState[3][1][1][3][3]` (gov_state)
- Previous parameters: `gs_previous_pparams` at `NewEpochState[3][1][1][3][4]` (gov_state)
- Future parameters: `gs_future_pparams` at `NewEpochState[3][1][1][3][5]` (gov_state)
So we know they're there!

### Acceptance Criteria
- [ ] Parameters state initializes from snapshot `gov_state` during bootstrap
- [ ] Current and previous protocol parameters loaded from snapshot
- [ ] Bootstrap data validates against Blockfrost parameters endpoints

### Testing Requirements
- [ ] Blockfrost `/epochs/{epoch}/parameters` endpoint matches bootstrap data

### Definition of Done
- Parameters state initializes from snapshot without replaying governance outcomes
- Current and previous parameters accurate after bootstrap
- Blockfrost API validation confirms correctness

lowhung:
@buddhisthead I'm moving this ticket to **Blocked** as we need to parse the Protocol Params I've mentioned in this ticket from the [NewEpochState cbor file](https://github.com/rrruko/nes-cddl-hs/blob/main/nes.cddl).

PR #521, opened 2025-12-18
refactor: only pass future param set to parameters_state module

whankinsiv:
## Description
This PR refactors the `GovernanceProtocolParametersBootstrapMessage` to publish only the single parameter set required for bootstrapping `parameter_state`, instead of publishing multiple parameter variants and deriving the active set inside the state module.

As part of this, the snapshot bootstrap logic now decodes directly into `ProtocolParamUpdate`, removing the snapshot specific conversion logic from `parameter_state` and simplifying the bootstrap path. 

## Related Issue(s)
Relates to #387

## How was this tested?
* Validated that parameters bootstrapped in `parameter_state` for epoch 507 match Blockfrost's response (with the exception of some missing  Plutus V3 cost models).
* Verified that `parameter_state` now publishes a single `ProtocolParamsMessage`, eliminating one potential source of deadlock during bootstrap. 

## Checklist

- [x] CI is green for this PR

## Impact / Side effects
* Parameters are now decoded directly into `ProtocolParamUpdate`, with current and future parameters merged before publishing the bootstrap message.
* Only the minimal set of previous/current parameters required for rewards calculation are retained in state via the new `RewardParams` type. 

Snapshot decoding changes in `/common/src/snapshot/protocol_parameters.rs`. 

Note: This refactor also makes it explicit which parameter set applies to a given epoch. I suspect part of our existing reward calculation issues stem from using the current parameter set rather than the parameter set that was active for the epoch being rewarded.

### DIFF
--- a/processes/omnibus/omnibus.toml
+++ b/processes/omnibus/omnibus.toml
@@ -5,7 +5,7 @@
 # ============================================================================
 [global.startup]
 network = "mainnet"
-method = "snapshot"  # Options: "mithril" | "snapshot"
+method = "mithril"  # Options: "mithril" | "snapshot"
 topic = "cardano.sequence.start"
 
 # ============================================================================
@@ -216,7 +216,3 @@ class = "in-memory"
 [[message-router.route]]  # Everything is internal only
 pattern = "#"
 bus = "internal"
-
-[monitor]
-output = "monitor.json" # Write to a file named monitor.json
-frequency_secs = 5.0 


common/examples/test_streaming_parser.rs
```diff
@@ -200,106 +199,48 @@ impl GovernanceProtocolParametersCallback for CountingCallbacks {
...
-        eprintln!("  Min Fee A: {}", gs_previous_params.min_fee_a);
-        eprintln!("  Min Fee B: {}", gs_previous_params.min_fee_b);
         eprintln!(
-            "  Max Block Body Size: {}",
-            gs_previous_params.max_block_body_size
+            "  Desired number of pools: {}",
+            previous_reward_params.desired_number_of_stake_pools
         );
         eprintln!(
-            "  Max Transaction Size: {}",
```
buddhisthead: why remove all this output from the test streamer? I use it.

common/examples/test_streaming_parser.rs
```diff
@@ -200,106 +199,48 @@ impl GovernanceProtocolParametersCallback for CountingCallbacks {
...
+            "  Pledge influence: {}",
+            current_reward_params.pool_pledge_influence
         );
+        eprintln!("  Min pool cost: {}", current_reward_params.min_pool_cost);
 
         // Store for later display
-        self.gs_previous_params = Some(gs_previous_params);
-        self.gs_current_params = Some(gs_current_params);
-        self.gs_future_params = Some(gs_future_params);
+        self.previous_reward_params = Some(previous_reward_params);
```
buddhisthead: are these names really better?

modules/snapshot_bootstrapper/src/publisher.rs
```diff
@@ -395,21 +388,16 @@ impl GovernanceProtocolParametersCallback for SnapshotPublisher {
...
-    );
+    info!("Received governance protocol parameters for epoch {epoch}",);
     // Send a message to the protocol parameters state, one per slice
     let message = Arc::new(Message::Snapshot(SnapshotMessage::Bootstrap(
         SnapshotStateMessage::ParametersState(GovernanceProtocolParametersBootstrapMessage {
             epoch,
-            slice,
             params,
-            era: Some(Era::Conway),
+            era: Era::Conway,
```
lowhung: You should be able to tack on the era (and maybe the network 🤷🏻) into the `EpochContext` struct, and pass those down into this `fn publish_gov_state(...)` method from the publisher itself. That way these don't need to be hardcoded.

Issue #499, opened 2025-12-15
Implement `UTxOW` rules for post `Shelley` eras


golddydev:
## Story

Implement `UTxOW` rules for post `Shelley` eras.

## Description

We can use almost same logic to check  `UTxOW` rules for post-`Shelley` eras.

## Acceptance Criteria

- [ ] Validate all transactions regardless of `Era`.

PR #528, opened 2025-12-23
feat: implement `Shelley`'s `UTxOW` rules for post-`Shelley` eras

golddydev:
## Description

This PR implements `Shelley`'s `UTxOW` rules for post-`Shelley` eras since they share almost same rules.

This is only for `tx_unpacker`, post-`Shelley` eras `UTxOW` rules for `utxos_state` will be implemented in coming PR.

## Related Issue(s)
Relates #499 

## How was this tested?
Added unit test cases for `Allegra` and `Conway` era.

## Checklist

- [x] CI is green for this PR

## Impact / Side effects
Now, it validates all transactions instead of `Shelley` era's transactions.

`tx_unpacker/validations/{era}/utxow.rs`

## NOTE:

They just wrap `Shelley`'s `validate` function. 
And newly introduced rules for specific Eras will be implemented in coming PR. (Added `NOTE::` for reminder)


modules/tx_unpacker/src/validations/mod.rs
```diff
@@ -1,52 +1,217 @@
...
 mod shelley;
 
-pub fn validate_shelley_tx(
+pub fn validate_tx(
     raw_tx: &[u8],
     genesis_delegs: &GenesisDelegates,
-    shelley_params: &ShelleyParams,
-    current_slot: u64,
+    shelley_params: &Option<ShelleyParams>,
+    block_info: &BlockInfo,
```
lowhung: Do we plan to use more of the `BlockInfo` in future validations? Otherwise it might be cleaner to just pass the `era` and `slot` to the `validate_tx` method, and it would be easier to test as well.

modules/tx_unpacker/src/validations/mary/utxow.rs
```diff
@@ -0,0 +1,30 @@
...
+        mtx,
+        tx_hash,
+        vkey_witnesses,
+        native_scripts,
+        genesis_delegs,
+        update_quorum,
+    )?;
+
+    Ok(())
+}
```
whankinsiv: If Mary and Allegra have the exact same validation rules as Shelley why not just group Shelley-Mary in the validation match arm instead of duplicating this logic?

modules/tx_unpacker/src/validations/mod.rs
golddydev: 👍🏼 
[02b6ab1](https://github.com/input-output-hk/acropolis/pull/528/commits/02b6ab1774a03b2ab34409e8531d6d9df9747730)

lowhung (approved):

Issue #402, opened 2025-11-21
Modify drep_state to report errors into validation channel


shd:
See https://github.com/input-output-hk/acropolis/issues/379 for details.

Report errors to 'cardano.validation.drep'

PR #529, opened 2025-12-25
Shd/drep validation outcomes

shd:
## Description

Converts error! to ValidationOutcome messages. Some effort is made to reduce boilerplate code

## Related Issue(s)
Fixes ticket #402 

## How was this tested?
- Unit tests
- Running node (replaying the whole blockchain)
- Manual tests

## Checklist

- [x] CI is green for this PR

## Impact / Side effects
Includes some refactoring, attention needed.

### DIFF
--- a/common/src/caryatid.rs
+++ b/common/src/caryatid.rs
@@ -80,12 +80,11 @@ pub enum RollbackWrapper<T> {
 }
 
 /// Declares locally tailored cardano reader struct, providing a lightweight wrapper around
-/// Subscribers from topics.
-/// Main intention is get rid of boilerplate code:
-/// (a) topic declaration, config param reading, reader initialization
-/// (b) reading from the topic and retriving data, attached to Cardano::`msg_constructor`
-/// constructor; all other constructors are ignored, rollbacks are processed according to
-/// function.
+/// Subscribers from topics. The Main intention is to get rid of boilerplate code, and simplify:
+/// (a) topic configuration, config parameters reading and reader initialization;
+/// (b) data reading from the topic subscriber. The data are taken from the enum constructor
+/// by the functions, provided in the struct being declared, so that the user does not need to
+/// manually specify pattern matching code.
 #[macro_export]
 macro_rules! declare_cardano_reader {
     ($reader_name:ident, $param:expr, $def_topic:expr, $msg_constructor:ident, $msg_type:ty) => {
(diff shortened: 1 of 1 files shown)


common/src/caryatid.rs
```diff
@@ -87,3 +89,190 @@ macro_rules! declare_cardano_reader {
...
 }
+
+#[derive(Debug)]
+pub enum RollbackWrapper<T> {
+    Rollback(Arc<Message>),
+    Normal((Arc<BlockInfo>, Arc<T>)),
+}
+
+#[macro_export]
+macro_rules! declare_cardano_rdr {
```
sandtreader: I think with common things like this - particularly macros, which can be hard to understand, we should avoid contractions like 'rdr' and it needs comments as to how to use it

common/src/caryatid.rs
```diff
@@ -87,3 +89,190 @@ macro_rules! declare_cardano_reader {
...
+                cfg: &Arc<Config>,
+            ) -> Result<Option<Self>> {
+                if do_create {
+                    Ok(Some(Self::new(ctx, cfg).await?))
+                } else {
+                    Ok(None)
+                }
+            }
+
+            pub async fn read_rb(&mut self) -> Result<RollbackWrapper<$msg_type>> {
```
sandtreader: read_with_rollback() ?

common/src/caryatid.rs
```diff
@@ -87,3 +89,190 @@ macro_rules! declare_cardano_reader {
...
+                    ),
+                    Message::Cardano((
+                        _blk,
+                        CardanoMessage::StateTransition(StateTransitionMessage::Rollback(_)),
+                    )) => Ok(RollbackWrapper::Rollback(res.clone())),
+                    msg => bail!("Unexpected message {msg:?} for {}", $param),
+                }
+            }
+
+            pub async fn read(&mut self) -> Result<(Arc<BlockInfo>, Arc<$msg_type>)> {
```
sandtreader: read_ignoring_rollbacks() ?

modules/drep_state/src/drep_state.rs
```diff
@@ -60,6 +86,13 @@ const DEFAULT_STORE_VOTES: (&str, bool) = ("store-votes", false);
 )]
 pub struct DRepState;
 
+struct DRepSubscriptions {
+    snapshot: Option<Box<dyn Subscription<Message>>>,
```
sandtreader: If snapshot can't use the macro, maybe explain why?

sandtreader (changes requested):
This is a really nice pattern - only comment is that for (hopefully) widely-used code like
this I'd prefer more descriptive naming and usage comments - which you have for some of it, just not all.

modules/drep_state/src/drep_state.rs
sandtreader: OK, I'm happy to leave this for now

sandtreader (approved):
Thanks for the improved docs!

PR #531, opened 2025-12-27
An approach to bootstrapping the block kes validator with op-cert counters

buddhisthead:
## Description

On the note of bootstrapping the op-cert counters…
* we have to get them from a block; they aren’t part of the NewEpochState snapshot
* we want the values of the last block of the bootstrapped epoch (e.g. last block of 507) the BlockContext can be extended to contain the op-cert counters and they can be forwarded to the block_kes_validator
* but the block cbor data we have only contains a single op-cert PoolId; there should be dozens of them

This only results in a single counter value. Instead, we probably want to load from the last block, which would contain all of the op-cert counters we need to bootstrap the kes validator module. Boot from the last block. @Matt what are the implications of this? It may mess up the point, etc.

## Related Issue(s)
TBD to make a ticket for this

## How was this tested?
I've run this in the debugger with breakpoints to validate the correct operation of the bootstrapping messages and initialization of the opcert counters. It also has a println to stderr for the experiment.

## Checklist

- [ ] My code builds and passes local tests
- [ ] CI is green for this PR

## Impact / Side effects
Replacing the block.cbor data of 507 with the last block of the epoch may affect other parts of the bootstrapping process, as yet to be discovered.

Let's discuss the problems associated with using the last block of the epoch as opposed to the first block.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,7 +1,6 @@
 {
   "version": "0.2.0",
   "configurations": [
-  
     {
         "type": "lldb",
         "request": "launch",
@@ -49,6 +48,32 @@
       },
       "args": [],
       "cwd": "${workspaceFolder}/processes/omnibus"
+    },
+    {
+      "type": "lldb",
+      "request": "launch",
+      "name": "Debug executable 'acropolis_process_omnibus bootstrap'",
+      "cargo": {
+        "args": [
+          "build",
+          "--bin=acropolis_process_omnibus",
+          "--package=acropolis_process_omnibus"
+        ],
+        "filter": {
+          "name": "acropolis_process_omnibus",
+          "kind": "bin"
+        }
+      },
+      "args": [
+        "--config",
+        "omnibus.toml",
+        "--config",
+        "omnibus.bootstrap.toml"
+      ],
+      "cwd": "${workspaceFolder}/processes/omnibus",
+      "env": {
+        "RUST_LOG": "info"
+      }
     },
     {
         "type": "lldb",
--- a/common/src/messages.rs
+++ b/common/src/messages.rs
@@ -509,6 +509,13 @@ pub struct SPOBootstrapMessage {
     pub spo_state: SPOState,
 }
 
+/// KES state bootstrap message containing opcert counters
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct KESStateBootstrapMessage {
+    pub block_number: u64,
+    pub opcert_counters: HashMap<PoolId, u64>,
+}
+
 #[allow(clippy::large_enum_variant)]
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 pub enum SnapshotStateMessage {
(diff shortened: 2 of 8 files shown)


buddhisthead:
We had some nice discussion with @lowhung and @whankinsiv, which I'll capture here. The right solution is to obtain these op-cert counters a different way and load them along with the snapshot. Here is the gist of the conversation:

Matt
  [Saturday at 2:44 PM](https://input-output-rnd.slack.com/archives/C08KZ8XR6FR/p1766875476022219?thread_ts=1766863263.907179&cid=C08KZ8XR6FR)
Hi Chris, re: option 1, we do use the last block of 507 already when bootstrapping 507. That's the block cbor you're seeing in the data/mainnet/blocks directory.
[2:45](https://input-output-rnd.slack.com/archives/C08KZ8XR6FR/p1766875542034809?thread_ts=1766863263.907179&cid=C08KZ8XR6FR)
https://cardanoscan.io/block/10802133
https://cardanoscan.io/blocks?epoch=507
Matt
  [Saturday at 2:57 PM](https://input-output-rnd.slack.com/archives/C08KZ8XR6FR/p1766876272761669?thread_ts=1766863263.907179&cid=C08KZ8XR6FR)
but the block cbor data we have only contains a single op-cert PoolId; there should be dozens of them
This isn't true sadly. This is cbor for a single block (the last block of 507), so we would expect to see a single op cert, the op cert of whichever pool produced that block. If we need all of the op certs / counters we'd need to replay all of the blocks for that epoch. We could export all of the cbor for blocks in the epoch from Mumak (demeter service) or we use Ogmios, as long as it has an endpoint to actually get that information.
Will
  [Saturday at 3:44 PM](https://input-output-rnd.slack.com/archives/C08KZ8XR6FR/p1766879083400849?thread_ts=1766863263.907179&cid=C08KZ8XR6FR)
Replaying the blocks for one epoch will not be sufficient as not all pools produce a block in an epoch. From my understanding we either need to replay all block headers from genesis or provide a bootstrap file which maps pools to their op cert counter value for the snapshot epoch.
Matt
  [Saturday at 4:11 PM](https://input-output-rnd.slack.com/archives/C08KZ8XR6FR/p1766880705302779?thread_ts=1766863263.907179&cid=C08KZ8XR6FR)
We could use [ledgerState/operationalCertificates](https://ogmios.dev/api/#operation-publish-/?QueryLedgerStateOperationalCertificates) in Ogmios (edited) 
ogmios.dev
[WebSocket API Reference (v6.14.*) - Ogmios](https://ogmios.dev/api/#operation-publish-/?QueryLedgerStateOperationalCertificates)
User manual for Ogmios, a lightweight protocol translation service for Cardano.
Paul Clark
  [Yesterday at 1:24 AM](https://input-output-rnd.slack.com/archives/C08KZ8XR6FR/p1766913881972319?thread_ts=1766863263.907179&cid=C08KZ8XR6FR)
Are the OpCert counters actually critical for validation?  Could we not just capture them the first time we see one for each SPO?
christopher.​tilt
  [3 minutes ago](https://input-output-rnd.slack.com/archives/C08KZ8XR6FR/p1767019583881779?thread_ts=1766863263.907179&cid=C08KZ8XR6FR)
[@Matt](https://input-output-rnd.slack.com/team/U09K0BXJB62) and [@Will](https://input-output-rnd.slack.com/team/U0924PXC73N) thank you, that helps me understand the role of blocks in the epoch! According to Pi, the Amaru team is planning to use Ogmios to get these values. So that jives with your suggestion, Matt.
Also, this check (of comparing expected op-certs against actual) is the only check that Amaru is not currently doing. We could skate by without this check for now and declare the 5.* milestones complete; then add a ticket for this last phase of boot-from-snapshot.

Issue #525, opened 2025-12-22
Unblock queue into later stage of epoch 508


lowhung:
### Context

The Consensus / validator modules support validation of blocks as they are consumed from upstream peers. When the Acropolis node bootstraps from a snapshot at epoch 507, and transitions into epoch 508, from 508 onwards it needs to properly validate incoming blocks against the correct consensus state for the new epoch (I think?). This works up to a point, then the whole process stalls/deadlocks after some blocks are processed.

### Problem

The message bus seems to stall at a certain point (actually once it's reached around the queue size of processed blocks (see omnibus.toml for details). I'm not entirely sure why this is happening yet.

1) Given you run `make run-bootstrap`
2) You'll see the following logs:
```log
Snapshot epoch=508 treasury=1537364145455061 reserves=7789863747110689 rewards=739886917312354 deposits=4373648000000 total_stake=22714109988244409 spos=3094 blocks=21025

ERROR: Verification mismatch: reserves for epoch=509 calculated=7789863747110689 desired=7789809997399209 difference=-53749711480
ERROR: Verification mismatch: treasury for epoch=509 calculated=1537364145455061 desired=1537376975934454 difference=12830479393

Monetary: eta=0.9734 rho=3/1000 monetary_expansion=22747484067083
Treasury: total_rewards=22840586933581 cut=1/5 increase=4568117386716 stake_rewards=18272469546865
After monetary change epoch=509 reserves=7785388732590471 treasury=1541932262841777
```

### Acceptance Criteria

- [x] Investigate block validation state after epoch 507→508 transition
- [x] Determine root cause of block
- [x] Rewards calculations / SPDD are correct for 508 (508 -> 509)

### Testing Requirements
- [x] Ensure node can proceed from 508 -> 509

### Definition of Done
- Processing blocks from peers proceeds normally and we can transition into 509

PR #535, opened 2025-12-30
feat: use Mithril after snapshot bootstrap completion

whankinsiv:
## Description
This PR refactors snapshot startup so that it publishes a `StartMithril` sync command, enabling fast block replay via `mithril_snapshot_fetcher` from epoch 508 to near tip. 

New Flow:
* `snapshot_bootstrapper` publishes a `StartMithril` sync command after bootstrapping. 
* When running in `snapshot` startup mode, `mithril_snapshot_fetcher` waits for the sync point, calls `fetch_blocks_from_point`, and begins publishing blocks. 
* Once all blocks from the Mithril snapshot have been replayed, `mithril_snapshot_fetcher` publishes a `FindIntersect` command. 
* `peer_network_interface` then resumes normal chain fetch, syncing from the last Mithril point to tip. 

## Related Issue(s)
Relates to #525 

## How was this tested?
* Verified that Mithril begins fetching upstream blocks at the correct point after bootstrapping. 

## Checklist

- [x] CI is green for this PR

## Impact / Side effects
Snapshot mode now fetches blocks via Mithril as the default startup flow.

### DIFF
--- a/modules/mithril_snapshot_fetcher/src/mithril_snapshot_fetcher.rs
+++ b/modules/mithril_snapshot_fetcher/src/mithril_snapshot_fetcher.rs
@@ -4,7 +4,7 @@
 use acropolis_codec::map_to_block_era;
 use acropolis_common::{
     commands::chain_sync::ChainSyncCommand,
-    configuration::{StartupMethod, StartupMode},
+    configuration::{StartupMode, SyncMode},
     genesis_values::GenesisValues,
     messages::{CardanoMessage, Command, Message, RawBlockMessage},
     BlockHash, BlockInfo, BlockIntent, BlockStatus, Point,
@@ -409,7 +409,7 @@ impl MithrilSnapshotFetcher {
 
     /// Main init function
     pub async fn init(&self, context: Arc<Context<Message>>, config: Arc<Config>) -> Result<()> {
-        if !StartupMethod::from_config(&config).is_mithril() {
+        if !SyncMode::from_config(&config).is_mithril() {
             return Ok(());
         }
 
--- a/modules/snapshot_bootstrapper/src/publisher.rs
+++ b/modules/snapshot_bootstrapper/src/publisher.rs
@@ -1,4 +1,4 @@
-use acropolis_common::configuration::StartupMethod;
+use acropolis_common::configuration::SyncMode;
 use acropolis_common::messages::SPOBootstrapMessage;
 use acropolis_common::MagicNumber;
 use acropolis_common::ProtocolParamUpdate;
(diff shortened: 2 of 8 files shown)


modules/mithril_snapshot_fetcher/src/mithril_snapshot_fetcher.rs
```diff
@@ -402,23 +403,23 @@ impl MithrilSnapshotFetcher {
...
             .unwrap_or(DEFAULT_BOOTSTRAPPED_SUBSCRIBE_TOPIC.1.to_string());
         info!("Creating subscriber for bootstrapped on '{bootstrapped_subscribe_topic}'");
+        let sync_command_topic = config
+            .get_string(DEFAULT_SYNC_COMMAND_TOPIC.0)
+            .unwrap_or(DEFAULT_SYNC_COMMAND_TOPIC.1.to_string());
+        info!("Publishing completion on '{sync_command_topic}'");
 
         let mut bootstrapped_subscription =
             context.subscribe(&bootstrapped_subscribe_topic).await?;
+        let mut sync_command_subscription = if StartupMethod::from_config(&config).is_snapshot() {
```
lowhung: Much cleaner this way 👍🏻

modules/snapshot_bootstrapper/src/publisher.rs
```diff
@@ -157,7 +157,7 @@ impl SnapshotPublisher {
             self.sync_command_topic,
             point.slot()
         );
-        let message = Message::Command(Command::ChainSync(ChainSyncCommand::FindIntersect(point)));
+        let message = Message::Command(Command::ChainSync(ChainSyncCommand::StartMithril(point)));
```
lowhung: Perhaps this is something we _could_ configure in the future? But I don't see a reason to not use the faster approach as the default for now. Nicely done!

processes/omnibus/omnibus.bootstrap.toml
```diff
@@ -7,7 +7,8 @@
 # ============================================================================
 [global.startup]
 network = "mainnet"
-method = "snapshot"  # Options: "mithril" | "snapshot"
+sync-mode = "snapshot"   # Options: "genesis" | "snapshot"
+sync-method = "mithril" # Options: "mithril" | "upstream"
```
buddhisthead: should this be "sync-mode"? It's inconsistent with the omnibus.toml file below which uses "sync-mode".

buddhisthead:
Can you confirm that this PR allows both switches to be set either way and they can mix and match? i.e 4 possible combinations?

startup-mode: genesis or snapshot
sync-mode: mithril or upstream

processes/omnibus/omnibus.bootstrap.toml
```diff
@@ -7,7 +7,8 @@
 # ============================================================================
 [global.startup]
 network = "mainnet"
-method = "snapshot"  # Options: "mithril" | "snapshot"
+sync-mode = "snapshot"   # Options: "genesis" | "snapshot"
```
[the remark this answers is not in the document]
lowhung: Also fixed in my PR 👍🏻

PR #536, opened 2025-12-30
feat: bootstrap VRF validator module

whankinsiv:
## Description
This PR adds snapshot bootstrapping support to the `block_vrf_validator` module.

The VRF validator now initializes its state from snapshot data by reusing the `AccountsBootstrapMessage` to derive:
* The active SPO set from the last epoch
* The active stake distribution from the last epoch. 

In addition, the previous `nonces.json` format has been replaced with a representation that matches the nonce structure stored in `epoch_state`.

## Related Issue(s)
Relates to #525 

## How was this tested?
* Verified that the nonce published by `epoch_state` matches the nonce reported by DB sync for VRF validation in epoch 508.
* Verified that all VRF validation errors are eliminated through epoch 508 when running in snapshot mode.
* Verified that Mithril startup mode continues to operate as expected. 

## Checklist

- [x] CI is green for this PR

## Impact / Side effects
* VRF validation now works correctly when bootstrapping from snapshots. 
* `nonces.json` format has changed to align with the nonce structure stored in `epoch_state`. 

Conversion logic from `AccountsBootstrapMessage` into VRF validator `Snapshot` type.

### DIFF
--- a/modules/block_vrf_validator/src/block_vrf_validator.rs
+++ b/modules/block_vrf_validator/src/block_vrf_validator.rs
@@ -74,16 +74,8 @@ impl BlockVrfValidator {
     /// Wait for and process snapshot bootstrap messages
     async fn wait_for_bootstrap(
         history: Arc<Mutex<StateHistory<State>>>,
-        mut snapshot_subscription: Option<Box<dyn Subscription<Message>>>,
+        mut snapshot_subscription: Box<dyn Subscription<Message>>,
     ) -> Result<()> {
-        let snapshot_subscription = match snapshot_subscription.as_mut() {
-            Some(sub) => sub,
-            None => {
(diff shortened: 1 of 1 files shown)


modules/block_vrf_validator/src/state.rs
```diff
@@ -149,4 +151,10 @@ impl State {
         };
         result.map_err(|e| Box::new((*e).into()))
     }
+
+    pub fn bootstrap(&mut self, vrf_data: AccountsBootstrapMessage) -> Result<()> {
```
lowhung: Given this works I'm down to go with it, but I'm just wondering why we use the same `AccountsBootstrapMessage` here, and not a distinct message type? Given we don't use the entirety of that struct, just the `mark` snapshot and the `pools`.

Issue #431, opened 2025-12-01
Implement `ValueNotConservedUTxO` rule on `utxos_state`


golddydev:
## Story
Most of transaction validation (`UTxO` rules are in `tx_unpacker` module), but for `ValueNotConservedUTxO` rule, we couldn't validate this rule in `tx_unpacker` because we don't know anything about UTxO value inside this module.

## Description
We can validate this rule inside `utxos_state` module though, with some additional transaction information (transaction fee, deposits, rewards, refunds)

We can tweak `UTXODeltasMessage` to include those additional information.

## Acceptance Criteria

- [x] Implement validate function for `ValueNotConservedUTxO` rule
- [x] Add test cases for this rule with transaction cbor
- [x] Integrate this check into `omnibus` process

PR #566, opened 2026-01-07
feat: implement `ValueNotConservedUTxO` rule

golddydev:
## Description

This PR implements `ValueNotConservedUTxO` `UTxO` rule check.

This PR introduces 2 new messages `StakeCertificatesDeltasMessage` and `PoolCertificatesDeltasMessage` which are to deliver information of how Stake addresses and pools registration/deregistration result in (e.g. is it newly registered, or already registered? for stake addresses how many lovelace is deposited/refunded...).

These 2 messages are published from `accounts_state`, `spo_state` each (since they have Stake Addresses and Pools state) and `utxo_state` will wait for those 2 messages to use them to validate transaction's balance. (Value Consumed == Value Produced)

## Related Issue(s)
#431

## How was this tested?
- Added simple one unit test case.
- Have run `omnibus` process and check there is no `ValueNotConservedUTxO` error on mainnet.

## Checklist

- [x] CI is green for this PR

## Impact / Side effects
Since `utxo_state` is waiting for 2 new messages, `utxo_state` will process a bit slower.
And also `accounts_state` and `spo_state` need to publish new message.

`accounts_state/certificates_deltas_publisher.rs`: Publishes `StakeCertificatesDeltasMessage`
`spo_state/certificates_deltas_publisher.rs`: Publishes `PoolCertificatesDeltasMessage`


whankinsiv (commented):
Could you please explain why `StakeCertificatesDeltasMessage ` and `PoolCertificatesDeltasMessage ` are needed? 

In cases where a registration or deregistration is invalid per ledger rules, it seems cleaner and more efficient for `accounts_state` / `spo_state` to reject the transaction via consensus.

golddydev:
> Could you please explain why `StakeCertificatesDeltasMessage ` and `PoolCertificatesDeltasMessage ` are needed?
> 
> In cases where a registration or deregistration is invalid per ledger rules, it seems cleaner and more efficient for `accounts_state` / `spo_state` to reject the transaction via consensus.

Okay. I thought that way, but to me every module should reject invalid transactions by itself, but this is overcomplicating things.
I'll remove those new Messages and rely on consensus to reject invalid transactions.
But still we need to have SPO state since `re-registration` of SPO is valid transaction.

modules/accounts_state/src/state.rs
```diff
@@ -881,25 +887,36 @@ impl State {
             };
 
             self.pots.deposits += deposit;
-        }
+            StakeCertificateOutcome::Registered(deposit)
+        } else {
+            // Already registered, no deposit taken
+            StakeCertificateOutcome::AlreadyRegistered
+        };
```
whankinsiv: This should be a validation failure.

golddydev:
Here is one more thing I forgot to consider, the mint and burn value.
I'll update those and request the review.

common/src/types.rs
```diff
@@ -661,6 +793,59 @@ impl Value {
     pub fn coin(&self) -> u64 {
         self.lovelace
     }
+
+    pub fn try_apply_assets_delta(&mut self, delta: &NativeAssetsDelta) -> Result<()> {
```
sandtreader: Not sure where (if at all?) this is used but it doesn't belong in types.rs I don't think.

modules/accounts_state/src/state.rs
sandtreader: Sorry I missed that update - great!

sandtreader (approved):
All good now, thanks for the updates!

PR #594, opened 2026-01-16
refactor: only store script hash in utxo_state

### LINKED ISSUE #577: Refactor utxo_state to store only script hash instead of full bytes
A new `script_store` or `script_state` module will be added to store the full script bytes, making storage in `utxo_state` redundant and wasteful of memory. Refactor to pass only the script hash to `utxo_state` and store this hash in `UTXOValue` to support the Blockfrost `/accounts/{stake_address}/utxos` endpoint.

whankinsiv:
## Description
This PR refactors `utxo_state` to store only script hashes instead of the full script bytes. 
* `tx_unpacker` now publishes the script hash in `TxUTxODeltasMessage`
* `utxo_state` now persists only the script hash

Full script bytes will be stored in `script_state`, keyed by script hash. I've also included reading the extra messages needed to maintain sync when in bootstrap mode. 

## Related Issue(s)
Closes #577 

## How was this tested?
* Verified /accounts/{stake_address}/utxos returns correct script hashes
* Verified /addresses/{address}/utxos returns correct script hashes

## Checklist
- [x] CI is green for this PR

## Impact / Side effects
Reduces memory usage in `utxo_state`.

N/A


modules/rest_blockfrost/src/types.rs
```diff
@@ -957,17 +955,7 @@ impl UTxOREST {
...
         };
 
-        let reference_script_hash = entry.reference_script.as_ref().map(|script| match script {
-            ReferenceScript::PlutusV1(b)
-            | ReferenceScript::PlutusV2(b)
-            | ReferenceScript::PlutusV3(b) => {
-                let mut hasher = Blake2b512::new();
-                hasher.update(b);
-                let result = hasher.finalize();
-                hex::encode(&result[..32])
```
lowhung: Why was it using 32 before vs the proper 28? And `Blake2b512` vs `Blake2bVar`?  Sorry, it's afternoon just trying to discern why these are different now (this map approach vs `hash_script_ref`)

lowhung (approved):
Looks good to me :)

modules/rest_blockfrost/src/types.rs
whankinsiv: We were incorrectly using `Blake2b512`. To generate the correct script reference hash we now use `Blake2b224` and include the Plutus version tag in the data being hashed. I've introduced `keyhash_224_tagged` in `/common/src/crypto.rs` for this purpose.

common/src/types.rs
```diff
@@ -523,6 +523,17 @@ pub enum ReferenceScript {
     PlutusV3(Vec<u8>),
 }
 
+impl ReferenceScript {
+    pub fn compute_hash(&self) -> Option<ScriptHash> {
+        match self {
+            ReferenceScript::Native(_) => None,
```
golddydev: We have `compute_hash` implemented for native scripts. Should we also compute hash for native scripts?

PR #605, opened 2026-01-20
fix: monetary difference / spdd verification 235 -> 506

### LINKED ISSUE #612: SPDD/Rewards: Verify Alonzo Era (epochs 290-364)
## Problem

Verify SPDD and rewards calculations are correct throughout the Alonzo era.

## Context

Alonzo introduced Plutus smart contracts and several staking-related changes:

- Script-based stake credentials
- Collateral inputs for script validation
- Potential for script-locked stake addresses

Key verification points:
- Script stake credentials receive rewards correctly
- Collateral handling doesn't affect stake calculations
- `is_pre_babbage` checks correctly include Alonzo for legacy bug compatibility (shared reward account bug, deregistration handling)

## Verification Steps

1. Enable in `processes/omnibus/omnibus.toml`:
   ```toml
   store-spdd = true
   ```
2. Run: `make run`
3. Once past epoch 364, run:
   ```bash
   cd tests/integration
   SPDD_VALIDATION_START_EPOCH=290 npx ts-node spdd.test.ts
   ```

## Expected Result

All epochs 290-364 should show 0 mismatches, 0 missing, 0 extra pools.

## Files

- `modules/accounts_state/src/rewards.rs`
- `common/src/epoch_snapshot.rs`

Part of #610

lowhung:
## Description

Implements AVVM (Ada Voucher Vending Machine) handling at the Allegra hard fork boundary, MIRs overwrite -> summation for pre-Alonzo & Alonzo, handles pointer address overflow, and generally corrects several reward calculation issues affecting epochs 235-506.

## Key changes

- **AVVM cancellation**: Dynamically cancel AVVM UTxOs at the Allegra boundary by querying `utxo_state` instead of hardcoding the value. The query handler performs cancellation on demand to avoid race conditions between modules.
- **MIR semantics**: Fix MIR (Move Instantaneous Rewards) override vs sum semantics — pre-Alonzo (epoch < 290) uses override (last value wins), Alonzo+ uses sum.
- **Pointer address overflow**: Resolve Pallas VarInt overflow in pointer address cache that caused panics on malformed pointer addresses.
- **Stability window from protocol params**: Replace hardcoded `STABILITY_WINDOW_SLOT` constant with computation from Shelley genesis params (`4k/f`).
- **Protocol param rotation**: Always rotate protocol params at epoch boundaries to fix k-parameter timing for rewards calculation.

## Related Issue(s)

Fixes #612, #611 and #609

## How was this tested?

- Ran full mainnet sync from genesis through epoch 506
- Verified SPDD matches db-sync reference data for epochs 208-506
- Verified rewards match db-sync reference data for epochs 208-506

## Checklist

- [x] CI is green for this PR

## Impact / Side effects

- UTxO state now handles AVVM cancellation query from accounts state (on-demand cancellation in query handler)
- Adds `cancel_redeem_utxos()` to all `ImmutableUTXOStore` implementations
- Adds `GetAvvmCancelledValue` query/response variants to UTxO state queries
- Adds `is_redeem_address()` / `is_redeem()` helpers to address types

- `utxo_state.rs:431-438` — On-demand AVVM cancellation in query handler
- `rewards.rs:177` — Era check change from Babbage to Allegra for shared reward account collision
- `rewards.rs:464` — Leader reward handling comment and pre-Babbage registration check
- `state.rs:425-427` — Stability window computed from Shelley genesis params (`4k/f`)
- `state.rs:345-368` — AVVM cancelled value query to utxo_state
- `stake_delta_filter/predefined.rs` — Pointer address cache overflow fix

### DIFF
--- a/common/src/address.rs
+++ b/common/src/address.rs
@@ -91,39 +91,13 @@ impl ByronAddress {
     /// - 0 = ATVerKey (regular verification key address)
     /// - 2 = ATRedeem (redeem/AVVM address)
     ///
+    /// The addr_type is the last element and CBOR-encodes small integers as
+    /// a single byte, so we can simply check if the last byte is 2.
+    ///
     /// At the Allegra hard fork (epoch 236 on mainnet), all redeem addresses
     /// are cancelled and their value returned to reserves.
     pub fn is_redeem_address(&self) -> bool {
-        // The payload is CBOR: [root_hash (28 bytes), attributes, addr_type]
-        // We need to decode it and check the third element
-        let mut decoder = minicbor::Decoder::new(&self.payload);
-
-        // Decode the array
-        let Ok(Some(len)) = decoder.array() else {
-            return false;
-        };
-
-        if len != 3 {
-            return false;
-        }
-
-        // Skip root_hash (bytes)
-        if decoder.bytes().is_err() {
-            return false;
-        }
-
-        // Skip attributes (map or other structure)
-        if decoder.skip().is_err() {
-            return false;
-        }
-
-        // Read addr_type - should be a u8
-        // ATVerKey = 0, ATRedeem = 2
-        let Ok(addr_type) = decoder.u8() else {
-            return false;
-        };
-
-        addr_type == 2
+        self.payload.last() == Some(&2)
     }
 }
 
--- a/modules/accounts_state/src/state.rs
+++ b/modules/accounts_state/src/state.rs
@@ -111,13 +111,15 @@ pub struct State {
     start_rewards_tx: Option<mpsc::Sender<()>>,
 
     /// Randomness stabilization window (4k/f slots), computed from protocol params.
-    /// Rewards calculation starts at this slot offset within the epoch.
-    /// In Cardano, `addrsRew` is captured at this point, so registration/deregistration
-    /// changes after the staking snapshot but before this slot affect reward eligibility.
+    /// Used to delay the rewards calculation and filter registration changes:
+    /// - `notify_block` waits until `block.epoch_slot >= stability_window_slot` before
+    ///   signalling the rewards task to start
+    /// - `apply_registration_changes_filtered` skips any `RegistrationChange` with
+    ///   `epoch_slot > stability_window_slot`, so only changes before this point
+    ///   affect `addrsRew` (the set of addresses eligible for rewards)
+    /// - Changes after this slot remain in `current_epoch_registration_changes` and
+    ///   are picked up at the next epoch boundary
     stability_window_slot: u64,
-
-    /// Flag to track if AVVM cancellation has been handled (happens once at Allegra boundary)
-    avvm_handled: bool,
 
     /// Pending MIRs from reserves to be applied at epoch boundary
     /// Key is stake address, value is the amount to add (or in Alonzo+, accumulated sum)
@@ -128,9 +130,6 @@ pub struct State {
     /// Key is stake address, value is the amount to add (or in Alonzo+, accumulated sum)
     /// Pre-Alonzo: last value wins (override). Alonzo+: values are summed.
     pending_mir_treasury: HashMap<StakeAddress, i64>,
-
-    /// Current era (set from BlockInfo on each block)
-    era: Era,
 }
 
 impl State {
@@ -377,6 +376,7 @@ impl State {
         &mut self,
         context: Arc<Context<Message>>,
         epoch: u64,
+        era: Era,
         is_new_era: bool,
         total_fees: u64,
         spo_block_counts: HashMap<PoolId, usize>,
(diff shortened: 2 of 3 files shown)


common/src/address.rs
```diff
@@ -83,6 +83,48 @@ impl ByronAddress {
...
+    /// Check if this Byron address is a redeem (AVVM) address.
+    ///
+    /// Byron address payload is CBOR: [root_hash, attributes, addr_type]
+    /// where addr_type is:
+    /// - 0 = ATVerKey (regular verification key address)
+    /// - 2 = ATRedeem (redeem/AVVM address)
+    ///
+    /// At the Allegra hard fork (epoch 236 on mainnet), all redeem addresses
+    /// are cancelled and their value returned to reserves.
+    pub fn is_redeem_address(&self) -> bool {
```
whankinsiv: Instead of decoding I believe you could use `self.payload.last() == Some(2)` which would be more efficient.

modules/accounts_state/src/state.rs
```diff
@@ -114,6 +109,28 @@ pub struct State {
...
+
+    /// Pending MIRs from reserves to be applied at epoch boundary
+    /// Key is stake address, value is the amount to add (or in Alonzo+, accumulated sum)
+    /// Pre-Alonzo: last value wins (override). Alonzo+: values are summed.
+    pending_mir_reserves: HashMap<StakeAddress, i64>,
+
+    /// Pending MIRs from treasury to be applied at epoch boundary
+    /// Key is stake address, value is the amount to add (or in Alonzo+, accumulated sum)
+    /// Pre-Alonzo: last value wins (override). Alonzo+: values are summed.
+    pending_mir_treasury: HashMap<StakeAddress, i64>,
```
whankinsiv: Same thing here.

modules/accounts_state/src/state.rs
```diff
@@ -114,6 +109,28 @@ pub struct State {
...
     start_rewards_tx: Option<mpsc::Sender<()>>,
+
+    /// Randomness stabilization window (4k/f slots), computed from protocol params.
+    /// Rewards calculation starts at this slot offset within the epoch.
+    /// In Cardano, `addrsRew` is captured at this point, so registration/deregistration
+    /// changes after the staking snapshot but before this slot affect reward eligibility.
+    stability_window_slot: u64,
+
+    /// Flag to track if AVVM cancellation has been handled (happens once at Allegra boundary)
+    avvm_handled: bool,
```
whankinsiv: I dont think we need this. We have `is_new_era` on `BlockInfo` so the check can be `if block_info.new_era && (block_info.era == Era::Allegra`.

modules/accounts_state/src/state.rs
```diff
@@ -324,24 +341,69 @@ impl State {
...
-        era: Era,
         is_new_era: bool,
         total_fees: u64,
         spo_block_counts: HashMap<PoolId, usize>,
         verifier: &Verifier,
     ) -> Result<Vec<StakeRewardDelta>> {
+        // At the Allegra hard fork boundary, all Byron redeem (AVVM) UTxOs are cancelled
+        // and their value returned to reserves. Query utxo_state for the cancelled amount,
+        // which it computes by scanning and removing all redeem-address UTxOs.
+        if self.era == Era::Allegra && !self.avvm_handled {
```
whankinsiv: the second condition can be `is_new_era`

modules/accounts_state/src/state.rs
```diff
@@ -1737,6 +1940,7 @@ mod tests {
         };
 
         state.pay_mir(&mir);
+        state.apply_pending_mirs(); // Apply accumulated MIRs
```
whankinsiv: This will cause underflows when an account attempts to withdrawal their MIR before the epoch boundary.

modules/accounts_state/src/state.rs
```diff
@@ -114,6 +109,28 @@ pub struct State {
 
     /// Signaller to start the above - delayed in early Shelley to replicate bug
     start_rewards_tx: Option<mpsc::Sender<()>>,
+
+    /// Randomness stabilization window (4k/f slots), computed from protocol params.
+    /// Rewards calculation starts at this slot offset within the epoch.
+    /// In Cardano, `addrsRew` is captured at this point, so registration/deregistration
+    /// changes after the staking snapshot but before this slot affect reward eligibility.
+    stability_window_slot: u64,
```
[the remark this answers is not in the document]
lowhung: They're still recorded in `current_epoch_registration_changes` — they just get filtered out when the rewards task runs. So the flow of it is...

1. At epoch boundary, `enter_epoch` spawns a rewards calculation task but blocks it on a channel (`start_rewards_rx`)
2. During the epoch, every registration/deregistration pushes a `RegistrationChange` (with its `epoch_slot`) into `current_epoch_registration_changes`
3. `notify_block` checks each block — once `block.epoch_slot >= stability_window_slot`, it sends the start signal
4. The rewards task wakes up, reads `current_epoch_registration_changes`, and calls `apply_registration_changes_filtered` with `Some(stability_window_slot)` — this skips any change where `epoch_slot > stability_window_slot`
5. Changes after the stability window aren't lost — they stay in `current_epoch_registration_changes` and get picked up at the next epoch boundary when the vec is drained into the snapshot's `registration_changes`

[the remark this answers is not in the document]
lowhung: Also updated my comment 8098d22c4a5257bea1552c797c0974ef74d01385

Issue #552, opened 2026-01-05
Investigate memory usage in full Data Node mode


sandtreader:
If you run Omnibus with all the data features turned on (see [this config](https://github.com/input-output-hk/acropolis/blob/prc/architecture/processes/omnibus/configs/ledger-with-api-and-history.toml), currently in prc/architecture only), it
climbs to 17GB by epoch 291 (I stopped it then) and outputs loads of Fjall errors - e.g.

```
2026-01-05T15:26:23.371085Z  INFO fjall::partition: partition: write halt because of write buffer saturation
```

others (lost in scrollback) about delaying 64us because L0 cache full etc.

These might be two different things - see also #551 - but needs profiling.

Issue #551, opened 2026-01-05
Move AssetState storage to disk


sandtreader:
AssetState is believed to take a lot of memory when it has the historical features turned on - if so, this needs to be shed to disk (Fjall) in the same way as AddressState etc.

PR #653, opened 2026-01-30
Ajw/639 improve asset state mem use

alexwoods:
## Description

This optimises the memory use for address state storage of assets. Relates to issue 

## Related Issue(s)
Relates to goal '639
Relates to issues #552 and #551.

## How was this tested?
Using a config that included just the modules that were enough to enable all the asset storage, and the memory profiling from PR #593, running up to epoch 300 and comparing usage.

## Checklist

- [x] CI is green for this PR

## Impact / Side effects
Functionality should remain the same.

Nothing in particular, although the rollback is the most fiddly bit so that might be worth some focus.

### DIFF
--- a/modules/assets_state/src/assets_state.rs
+++ b/modules/assets_state/src/assets_state.rs
@@ -54,7 +54,7 @@ pub struct AssetsState;
 impl AssetsState {
     async fn run(
         history: Arc<Mutex<StateHistory<State>>>,
-        address_state: Arc<Mutex<AddressState>>,
+        address_state: Option<Arc<Mutex<AddressState>>>,
         mut asset_deltas_subscription: Box<dyn Subscription<Message>>,
         mut utxo_deltas_subscription: Option<Box<dyn Subscription<Message>>>,
         mut address_deltas_subscription: Option<Box<dyn Subscription<Message>>>,
@@ -93,12 +93,14 @@ impl AssetsState {
                         let mut reg = registry.lock().await;
                         state = match state.handle_mint_deltas(&deltas_msg.deltas, &mut reg) {
                             Ok((new_state, updated_asset_ids)) => {
-                                let mut address_state = address_state.lock().await;
-                                address_state.new_block(
-                                    block_info.number,
-                                    &updated_asset_ids,
-                                    &reg,
-                                );
+                                if let Some(ref address_state) = address_state {
+                                    let mut address_state = address_state.lock().await;
+                                    address_state.new_block(
+                                        block_info.number,
+                                        &updated_asset_ids,
+                                        &reg,
+                                    );
+                                }
                                 new_state
                             }
                             Err(e) => {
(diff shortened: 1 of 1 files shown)


modules/assets_state/src/address_state.rs
```diff
@@ -0,0 +1,367 @@
...
+    params::SECURITY_PARAMETER_K, Address, AddressDelta, AssetAddressEntry, ShelleyAddress,
+};
+use anyhow::Result;
+use std::collections::{HashMap, VecDeque};
+use tracing::error;
+
+#[derive(Debug, PartialEq)]
+pub struct AddressState {
+    // Assets mapped to addresses
+    pub addresses: HashMap<AssetId, HashMap<ShelleyAddress, u64>>,
```
whankinsiv: Could you please explain how this fixes the memory usage issue? It appears that addresses are still fully kept in memory and have just moved to their own state.

modules/assets_state/src/address_state.rs
```diff
@@ -0,0 +1,367 @@
...
+
+    // History of new assets added in previous blocks
+    asset_history: VecDeque<Vec<AssetId>>,
+
+    // Last block number for detecting rollbacks
+    last_block_num: Option<u64>,
+}
+
+impl AddressState {
+    pub fn new() -> Self {
```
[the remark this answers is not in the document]
alexwoods: That's a good point. I'll have a look at that on Wednesday

modules/assets_state/src/state.rs
```diff
@@ -55,9 +54,6 @@ pub struct State {
     /// Assets mapped to extended info
     pub info: Option<HashMap<AssetId, AssetInfoRecord>>,
 
-    /// Assets mapped to addresses
-    pub addresses: Option<HashMap<AssetId, std::collections::HashMap<ShelleyAddress, u64>>>,
-
     /// Assets mapped to transactions
     pub transactions: Option<HashMap<AssetId, Vector<TxIdentifier>>>,
```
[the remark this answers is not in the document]
alexwoods: Agreed. All the bits that are past the volatile block (K 2160) that we need to server for blockfrost requests should be disk-based storage. The problem is finding a storage method that works well with the data. I'm not sure a key-value db is going to work well if there's an ever-growing transaction list to store

modules/assets_state/src/address_state.rs
whankinsiv: Wow! An almost 50% memory reduction is impressive for switching to `std::collections::HashMap`.

whankinsiv (approved):

modules/assets_state/src/assets_state.rs
```diff
@@ -164,12 +175,12 @@ impl AssetsState {
                         Self::check_sync(&current_block, block_info, "address");
 
                         let reg = registry.lock().await;
-                        state = match state.handle_address_deltas(&address_deltas_msg.deltas, &reg)
+                        let mut address_state = address_state.lock().await;
+                        match address_state.handle_address_deltas(&address_deltas_msg.deltas, &reg)
```
lowhung: Because `handle_address_deltas` returns `Result<()>`, this line `Ok(new_state) => new_state` binding is slightly misleading (we aren't returning any state here), so I think this could just be


```rust
if let Err(e) = address_state.handle_address_deltas(&address_deltas_msg.deltas, &reg) {
                  error!("Address deltas handling error: {e:#}");
              }
```

Issue #666, opened 2026-02-05
Validate Transaction Aux data


golddydev:
## Description

According to cardano soft fork, we need to validate transaction's `aux_data` after major version 2.

## Acceptance Criteria

- [x] Validate transaction's aux data is formed correctly

Issue #489, opened 2025-12-12
Implement `Metadata` validation


golddydev:
## Story

While implementing `UTxOW` rules, Tx `Metadata` check is missing.

## Description
**Rules**:
- [x] **MissingTxBodyMetadataHash** Tx has Metadata but no hash
- [x] **MissingTxMetadata** Tx has metadata hash, but no Metadata
- [x] **ConflictingMetadataHash** Metadata hash mismatch with its content
- [ ] **InvalidMetadata** (Now moved to #666 )

## Acceptance Criteria

- [ ] Add functions to check `Metadata` rules
- [ ] Add test cases with mock transaction
- [ ] Integrate to omnibus process

PR #667, opened 2026-02-05
feat: implement transaction's metadata validation

golddydev:
## Description

This PR implements basic transaction's metadata validations.

Except transaction's auxiliary data check. #666 

## Related Issue(s)
Related to #489 

## How was this tested?
* Added raw transaction cbor (which has metadata) for test case

## Checklist

- [x] CI is green for this PR

## Impact / Side effects
None

* `validate_metadata` function at `modules/tx_unpacker/src/validations/shelley/utxow.rs`
* `common/src/soft_fork`: Protocol version's soft fork.

### DIFF
--- a/common/src/soft_fork.rs
+++ b/common/src/soft_fork.rs
@@ -2,7 +2,7 @@ use crate::protocol_params::ProtocolVersion;
 
 /// We need to check transaction's auxiliary data since this soft-fork
 /// Which was activated after PV 2.0
-/// 
+///
 /// This checks metadata is correctly formed and their size limit.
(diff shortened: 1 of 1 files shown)


whankinsiv (approved):

lowhung (commented):
Ok one question cause I couldn't comment on inline since the file isn't in the diff `modules/tx_unpacker/src/validations/conway/utxow.rs` still has the old `// TODO: Validate metadata` comment and doesn't call `validate_metadata`. Since I presume Conway inherits the Shelley/Babbage rules, should this get wired up here too? Or is the plan to have Conway go through the Babbage validation path first?

Issue #698, opened 2026-02-17
Create scaffolding for `midnight-state` module


whankinsiv:
Create the general module layout with indexed state, message subscriptions, and placeholder subscription handlers based off the requirements outlined [here](https://docs.google.com/document/d/130r8tv14f_Hhy05uAkPsOIK8iuuBxQEPONueXIPxLbY/edit?tab=t.0).

PR #701, opened 2026-02-17
feat: midnight state module skeleton

whankinsiv:
## Description
This PR adds the skeleton for the `midnight_state` module which will be used to index and serve data to the `midnight-node` via gRPC. The only subscription this module needs is `AddressDeltasMessage`.

## Related Issue(s)
Relates to #698

## How was this tested?
* Verified that mainnet can still sync from genesis and booting from snapshot.

## Checklist

- [x] CI is green for this PR

## Impact / Side effects
Provides a starting place to continue work on the midnight indexer.

Types used in `State`.

### DIFF
--- a/modules/midnight_state/src/midnight_state.rs
+++ b/modules/midnight_state/src/midnight_state.rs
@@ -25,8 +25,6 @@ declare_cardano_reader!(
     AddressDeltasMessage
 );
 
-const DEFAULT_ENABLED_STATUS: (&str, bool) = ("enable-indexer", false);
-
 /// Midnight State module
 #[module(
     message_type(Message),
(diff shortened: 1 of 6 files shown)


modules/midnight_state/src/midnight_state.rs
```diff
@@ -0,0 +1,98 @@
...
+
+declare_cardano_reader!(
+    AddressDeltasReader,
+    "address-deltas-topic",
+    "cardano.address.deltas",
+    AddressDeltas,
+    AddressDeltasMessage
+);
+
+const DEFAULT_ENABLED_STATUS: (&str, bool) = ("enable-indexer", false);
```
lowhung: Is this necessary? I was operating under the assumption that including/omitting `[module.midnight-state]` from the `omnibus.toml` file would be sufficient for enabling/disabling respectively.

modules/midnight_state/src/types.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub tx_hash: TxHash,
+    pub utxo_tx_hash: TxHash,
+    pub utxo_index: u16,
+}
+
+/// ---------------------------------------------------------------------------
+/// Internal State Types
+/// ---------------------------------------------------------------------------
+///
+/// These structs are used internally by the indexing state and are not
```
lowhung: 👍🏻

processes/omnibus/omnibus-local.toml
```diff
@@ -22,6 +22,10 @@ node-addresses = [
 ]
 magic-number = 2
 
+[module.midnight-state]
```
lowhung: Should this be included in these configs? I think a bespoke toml config that includes the midnight indexer would be more appropriate than including this in all of these configs. We do have a lot of different configs admittedly...

Issue #697, opened 2026-02-17
Expand AddressDeltasMessage with optional extended info


whankinsiv:
Refactor `AddressDeltasMessage` as an enum:
```
pub enum AddressDeltasMessage {
    Deltas(AddressDeltas),
    ExtendedDeltas(ExtendedAddressDeltas)
}
```

where `ExtendedAddressDeltas` includes:
```
pub struct ExtendedAddressDeltas {
   // Address involved in delta
   pub address: Address,


   // Transaction in which delta occured
   pub tx_identifier: TxIdentifier,


   // Address impacted UTxOs
   pub spent_utxos: Vec<SpentUTxOExtended>,
   pub created_utxos: Vec<CreatedUTxOExtended>,


   // Sums of spent and created UTxOs
   pub sent: Value,
   pub received: Value,
}


pub struct SpentUTxOExtended {
   pub utxo: UTxOIdentifier,
   pub value: Value,
   pub spent_by: TxHash,
   pub datum: Option<Datum>,
}

pub struct CreatedUTxOExtended {
   pub utxo: UTxOIdentifier,
   pub value: Value,
   pub datum: Option<Datum>,
}
```

Issue #677, opened 2026-02-10
Add resolved UTxO delta publisher to utxo_state module


lowhung:
## Summary

Add a new message publisher to the `utxo_state` module that emits per-UTxO resolved data for both consumed and produced UTxOs in each transaction. This provides the data foundation for the Midnight indexer module.

## Problem

The existing `AddressDeltasMessage` aggregates UTxO changes per-address per-transaction, which loses:
- **Per-UTxO asset quantities** — if a tx produces 3 UTxOs at the same address, two holding cNight with different quantities, the `received: Value` sums them. Midnight needs per-UTxO quantity.
- **Datums on consumed inputs** — `spent_utxos: Vec<UTxOIdentifier>` gives only the identifier, not the datum that was on the consumed UTxO. Needed for deregistration events.
- **Per-UTxO value on consumed inputs** — the resolved `UTXOValue` (address, value, datum) for each consumed input is already computed in `handle_valid_tx` at `state.rs:516` but then flattened into aggregated sums.

### New Message Types

In `common/src/messages.rs`:

```rust
pub struct ResolvedUtxoDelta {
    pub utxo_id: UTxOIdentifier,    // (tx_hash, output_index)
    pub address: Address,
    pub value: Value,               // per-UTxO lovelace + native assets
    pub datum: Option<Datum>,
}

pub struct ResolvedTxDeltas {
    pub tx_identifier: TxIdentifier,  // (block_number, tx_index)
    pub consumed: Vec<ResolvedUtxoDelta>,
    pub produced: Vec<ResolvedUtxoDelta>,
}

pub struct ResolvedDeltasMessage {
    pub deltas: Vec<ResolvedTxDeltas>,
}
```

Add `CardanoMessage::ResolvedDeltas(ResolvedDeltasMessage)` variant.

### New Publisher

Implement `ResolvedDeltaPublisher` using the same observer pattern as the existing `AddressDeltaPublisher` (`modules/utxo_state/src/address_delta_publisher.rs`):
- Accumulate per-block resolved deltas
- Publish on `finalise_block`
- Handle rollbacks

### Integration

Wire into `State` in `utxo_state` as a new observer alongside the existing `AddressDeltaObserver`. The data is already computed in `handle_valid_tx`:
- For consumed inputs: `self.lookup_utxo(input)` already resolves the full `UTXOValue` (line 516)
- For produced outputs: `TxOutput` already has address, value, and datum

### Configuration

[...]

PR #702, opened 2026-02-18
feat: extend address deltas with extended payload

lowhung:
## Description
This PR refactors address delta publishing to support two payload shapes (`compact` and `extended`), default path uses `compact`.

`utxo_state` now publishes exactly one type per run based on `address-delta-publish-mode` (default `compact`, optional `extended` for midnight).

`midnight_state` logs concise epoch summaries from an internal observer-style totals accumulator.

## Related Issue(s)
- Relates to #697
- Relates to #677

## How was this tested?
  - Ran midnight configuration and confirmed epoch checkpoint logs are emitted by midnight state module
  - Normal config path runs correctly still

## Checklist
- [x] CI is green for this PR

## Impact / Side effects
So the main side-effects / impactful pieces are:
- `AddressDeltasMessage` is now an enum and downstream consumers must handle both variants. Existing compact consumers in this repo were updated via `as_compact_or_convert()`.
- **Configuration:** midnight deployments should use `address-delta-publish-mode = "extended"` in `module.utxo-state` while keeping the single address delta topic.

### DIFF
--- a/modules/midnight_state/src/state.rs
+++ b/modules/midnight_state/src/state.rs
@@ -1,18 +1,10 @@
 use anyhow::{anyhow, Result};
-use std::{
-    collections::{BTreeMap, HashMap},
-    sync::Arc,
-};
+use std::collections::{BTreeMap, HashMap};
 
-use acropolis_common::{
-    messages::AddressDeltasMessage, BlockInfo, BlockNumber, Datum, Epoch, UTxOIdentifier,
-};
+use acropolis_common::{messages::AddressDeltasMessage, BlockInfo, BlockNumber, Datum, Epoch};
 
 use crate::epoch_totals::{EpochSummary, EpochTotals};
-use crate::types::{
-    AssetCreate, AssetSpend, CandidateUTxO, Deregistration, DeregistrationEvent, Registration,
-    RegistrationEvent, UTxOMeta,
-};
+use crate::indexes::{candidate_state::CandidateState, cnight_utxo_state::CNightUTxOState};
 
 #[derive(Clone, Default)]
 pub struct State {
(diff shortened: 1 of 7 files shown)


common/src/types.rs
```diff
@@ -379,6 +380,53 @@ pub struct AddressDelta {
...
+
+    /// Full value of the spent UTxO
+    pub value: Value,
+
+    /// Hash of the transaction spending this UTxO
+    pub spent_by: TxHash,
+
+    /// Datum attached to the spent UTxO, if present
+    pub datum: Option<Datum>,
+}
```
whankinsiv: I don't think we need the Value nor the datum here as we already indexed these when the UTxO's were created.

modules/midnight_state/src/state.rs
```diff
@@ -4,15 +4,129 @@ use std::{
...
+                    extended_deltas.iter().map(|delta| delta.spent_utxos.len()).sum::<usize>();
+            }
+        }
+    }
+
+    fn finalise_block(&mut self, block: &BlockInfo) {
+        self.last_checkpoint = Some(EpochCheckpoint::from_block(block));
+    }
+}
+
```
whankinsiv: Could you move this to its own file please?

modules/utxo_state/src/state.rs
```diff
@@ -9,15 +9,45 @@ use acropolis_common::{
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "compact" => Ok(Self::Compact),
+            "extended" => Ok(Self::Extended),
+            _ => Err(anyhow::anyhow!(
+                "Invalid address-delta-publish-mode '{s}', expected 'compact' or 'extended'"
+            )),
+        }
+    }
+}
```
whankinsiv: maybe move this into `address_delta_publisher.rs`?

modules/utxo_state/src/state.rs
```diff
@@ -491,9 +528,23 @@ impl State {
...
             } else {
-                self.handle_invalid_tx(tx, block).await?;
+                match self.address_delta_publish_mode {
+                    AddressDeltaPublishMode::Compact => {
+                        self.handle_invalid_tx_compact(tx, block).await?
+                    }
+                    AddressDeltaPublishMode::Extended => {
+                        self.handle_invalid_tx_extended(tx, block).await?
+                    }
+                }
```
whankinsiv: nice!

PR #715, opened 2026-02-19
feat: add midnight process with subset of modules

whankinsiv:
## Description
Adds a new process: `acropolis_process_midnight` which includes only the needed modules for running the midnight data node. Updates the `make run-midnight` to run the midnight process instead of the omnibus process. 

## Related Issue(s)
N/A

## How was this tested?
* Verified that sync works as expected with the reduced set of modules.

## Checklist

- [x] CI is green for this PR

## Impact / Side effects
Running the midnight process initializes a smaller set of modules when compared to omnibus, reducing compile time and improving performance.

Giving `enact-state-topic` a default value in `parameter_state.rs` so that it does not need to be specified in the toml.


processes/midnight/acropolis.preview.toml
lowhung: What should happen to the old config toml for midnight?

We should probably update the docker processes to use these new files.

golddydev (approved):

PR #724, opened 2026-02-19
feat: use ValueMap for extended UTxO deltas

lowhung:
## What
- Switch extended address-delta value fields from Value to ValueMap
- Emit ValueMap directly in utxo_state extended mode instead of converting maps back into vectors
- Keep compact compatibility by converting extended ValueMap values back to Value only inside AddressDeltasMessage::as_compact_or_convert()
- Update tests accordingly and add local IDE/plan artifacts to .gitignore

## Why
Using ValueMap in extended deltas gives map-based asset lookups for downstream indexers (especially midnight), which makes asset filtering simpler and more efficient than repeatedly scanning vector-based Value.assets.

### DIFF
--- a/modules/utxo_state/src/state.rs
+++ b/modules/utxo_state/src/state.rs
@@ -1291,7 +1291,8 @@ mod tests {
                         delta.address,
                         Address::Byron(ByronAddress { ref payload }) if payload[0] == 99
                     ));
-                    let lovelace_net = (delta.received.lovelace as i64) - (delta.sent.lovelace as i64);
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -6,6 +6,9 @@ configuration/
 docker-compose.yaml
 .DS_Store
 
+# Jetbrains IDE files
+.idea
```
lowhung: I'd like to ignore the files staged automatically by Jetbrains IDEs, it's very noise-y for me 😓

PR #730, opened 2026-02-21
feat(midnight-state): process governance datums from address deltas

lowhung:
## Summary
- Index technical committee and council datums from extended address deltas
- Gate indexing by configured governance addresses and policy ids
- Add tests for matching/non-matching filters and rollback replay behavior

### DIFF
--- a/modules/midnight_state/src/state.rs
+++ b/modules/midnight_state/src/state.rs
@@ -3,7 +3,8 @@ use std::collections::HashSet;
 use anyhow::Result;
 
 use acropolis_common::{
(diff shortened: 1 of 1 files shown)


modules/midnight_state/src/state.rs
```diff
@@ -159,15 +160,48 @@ impl State {
...
+            return;
+        }
+
+        for created in &delta.created_utxos {
+            let Some(datum) = &created.datum else {
+                continue;
+            };
+
+            if is_technical_committee_address
+                && created.value.assets.contains_key(&self.config.technical_committee_policy_id)
```
whankinsiv: Nice!

PR #753, opened 2026-02-26
Pi/custom genesis

Quantumplation:
## Description

Adds support for custom genesis files, in place of well known network names. Also fixes a bug that this uncovered.

## Related Issue(s)
N/A

## How was this tested?
Pointed Acropolis (technically scooper-v2) at a [yaci-devkit](https://github.com/bloxbean/yaci-devkit) private devnet and confirmed it was able to sync from origin to tip.

## Checklist

- [X] My code builds and passes local tests
- [X] branch has ≤ 5 commits (honor system)
- [X] commit messages tell a coherent story
- [X] branch is up to date with main (rebased on main; fast-forward possible)
- [ ] CI/CD passes on the merged-with-main result

## Impact / Side effects
This will allow users using custom networks, like [yaci-devkit](https://github.com/bloxbean/yaci-devkit) to sync an acropolis node. It will allow users building custom tools like indexers to do the same.

### DIFF
--- a/modules/custom_indexer/src/index_actor.rs
+++ b/modules/custom_indexer/src/index_actor.rs
@@ -75,6 +75,17 @@ impl IndexActor {
     }
 
     pub async fn apply_txs(&mut self, block: Arc<BlockInfo>, txs: &[Arc<[u8]>]) {
+        // Origin is a virtual starting point (slot 0, no hash). Replace it with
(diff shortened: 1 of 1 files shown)


whankinsiv (approved):
Thanks for adding this Pi!

PR #769, opened 2026-03-04
snapshot bootstrap: load network genesis and praos params

lowhung:
This PR removes hardcoded Praos assumptions from snapshot bootstrap.

Snapshot bootstrap now publishes epoch bootstrap data without Praos params (`praos_params: None`) and no longer selects Praos params by network in bootstrap context. Epoch modules continue deriving Praos params from protocol parameters (`ShelleyParams -> PraosParams`) when protocol params are read.

This keeps bootstrap wiring simpler and avoids duplicating network-specific Praos constants in snapshot bootstrap.

### DIFF
--- a/modules/snapshot_bootstrapper/src/publisher.rs
+++ b/modules/snapshot_bootstrapper/src/publisher.rs
@@ -221,7 +221,6 @@ impl SnapshotPublisher {
             total_fees: data.total_fees_current,
             spo_blocks: data.spo_blocks_current.clone(),
             nonces: ctx.nonces.clone(),
-            praos_params: None,
         }
     }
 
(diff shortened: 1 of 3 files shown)


modules/snapshot_bootstrapper/src/publisher.rs
```diff
@@ -221,7 +224,7 @@ impl SnapshotPublisher {
             total_fees: data.total_fees_current,
             spo_blocks: data.spo_blocks_current.clone(),
             nonces: ctx.nonces.clone(),
-            praos_params: Some(PraosParams::mainnet()),
```
lowhung: Originally hard-coded here

PR #788, opened 2026-03-11
[fix] encode cnight grpc owner addresses

lowhung:
## Description

Fix cNIGHT gRPC owner-address encoding so Acropolis matches the shape `midnight-node` expects.

This change now normalizes cNIGHT holder addresses to stake/reward addresses at persistence time,
not in the gRPC layer. It also moves that conversion into `acropolis_common::Address::to_stake_address()`
so the logic is shared instead of being `midnight_state`-local.

Result:
- cNIGHT create/spend events serialize 29-byte owner stake/reward address bytes
- unsupported address shapes are skipped during ingest with warnings
- the internal field name `holder_address` stays unchanged

## Related Issue(s)

Relates to the preview gRPC sync failure where `midnight-node-grpc` rejected cNIGHT events with
`Invalid address length`.

## How was this tested?

- `cargo test -p acropolis_common -p acropolis_module_midnight_state`
- `cargo fmt --all --check`
- `cargo clippy -p acropolis_common -p acropolis_module_midnight_state --tests`

## Checklist

## Impact / Side effects

`AssetCreate.address` and `AssetSpend.address` now effectively represent 29-byte owner
stake/reward address bytes, which is what the current Midnight node client expects.

Please focus on:
- `common/src/address.rs`
- `modules/midnight_state/src/state.rs`
- `modules/midnight_state/src/types.rs`

### DIFF
--- a/modules/midnight_state/src/state.rs
+++ b/modules/midnight_state/src/state.rs
@@ -1,11 +1,10 @@
 use std::collections::HashSet;
 
-use anyhow::{anyhow, Result};
+use anyhow::Result;
 
 use acropolis_common::{
-    messages::AddressDeltasMessage, protocol_params::Nonce, Address, BlockInfo, BlockNumber, Datum,
-    Epoch, ExtendedAddressDelta, ShelleyAddressDelegationPart, StakeAddress, StakeCredential,
-    UTxOIdentifier,
+    messages::AddressDeltasMessage, protocol_params::Nonce, BlockInfo, BlockNumber, Datum, Epoch,
+    ExtendedAddressDelta, UTxOIdentifier,
 };
 use imbl::HashMap;
 use tracing::warn;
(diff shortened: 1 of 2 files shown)


modules/midnight_state/src/state.rs
```diff
@@ -365,6 +382,30 @@ impl State {
...
+            ShelleyAddressDelegationPart::None => {
+                Err(anyhow!("holder address has no delegation part"))
+            }
+        },
+        Address::Stake(stake) => Ok(stake.clone()),
+        Address::Byron(_) => Err(anyhow!("byron addresses are not supported")),
+        Address::None => Err(anyhow!("none addresses are not supported")),
+    }
+}
+
```
whankinsiv: I would add a `to_stake_address()` method on `Address`.

whankinsiv (approved):

Issue #811, opened 2026-03-24
Make modules rollback safe


whankinsiv:
Refactor all modules currently using `read_ignoring_rollbacks()` to instead use `read_with_rollbacks` so that message synchronization can stay aligned.

PR #812, opened 2026-03-24
refactor: add rollback support to drdd_state, spdd_state, and block_vrf_validator

whankinsiv:
## Description

This PR adds rollback safety to drdd_state, spdd_state, and block_vrf_validator by replacing their calls to `read_ignoring_rollbacks()` with `read_with_rollbacks()`. 

## Related Issue(s)
Relates to #811

## How was this tested?
* Verified that sync progresses as expected

## Checklist

## Impact / Side effects
The 3 refactored modules are now rollback safe and will stay aligned when a rollback message is received.

* Message read pattern in `block_vrf_validator`
* Moving `StateHistory` out of `State` for `spdd_state` and `drdd_state`

### DIFF
--- a/modules/block_vrf_validator/src/block_vrf_validator.rs
+++ b/modules/block_vrf_validator/src/block_vrf_validator.rs
@@ -12,7 +12,6 @@ use acropolis_common::{
     },
     protocol_params::Nonce,
     state_history::{StateHistory, StateHistoryStore},
-    validation::ValidationOutcomes,
     BlockStatus,
 };
 use anyhow::{bail, Result};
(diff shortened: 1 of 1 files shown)


modules/block_vrf_validator/src/block_vrf_validator.rs
```diff
@@ -137,125 +170,100 @@ impl BlockVrfValidator {
...
             // Get a mutable state
             let mut state = history.lock().await.get_or_init_with(State::new);
-            let mut current_block: Option<BlockInfo> = None;
 
-            let (_, message) = block_subscription.read_ignoring_rollbacks().await?;
-            match message.as_ref() {
-                Message::Cardano((block_info, CardanoMessage::BlockAvailable(block_msg))) => {
-                    // handle rollback here
+            let block_msg = match ctx
+                .consume_sync("block available", block_reader.read_with_rollbacks().await)?
```
lowhung: What is the convention for these handler strings? This feels [magic string-y] (https://deviq.com/antipatterns/magic-strings/) even though it looks like it's just for logging, I still think a convention here would be nice 🤔

Cause I've also seen 

```rust
match ctx.consume_sync("block_txs", txs_reader.read_with_rollbacks().await)? {
                RollbackWrapper::Normal((blk_info, msg)) => {
                    let span = info_span!("epochs_state.handle_block_txs", block = blk_info.number);
                    span.in_scope(|| state.handle_block_txs(&blk_info, &msg));
                }
                RollbackWrapper::Rollback(_) => {}
            }
``` 

Where we use `snake_case`

whankinsiv: I agree a standard convention would be nice. I think it may be easier to do a second pass on all the modules separately if that fine with you.

Issue #818, opened 2026-03-26
Security parameter should come from protocol params message


ChrisPaprocki:
Instead of using default value for security param as in:
* https://github.com/input-output-hk/acropolis/pull/709#discussion_r2865649359
* https://github.com/input-output-hk/acropolis/pull/789#discussion_r2989161675

it should be coming from the protocol params message security_param field.

buddhisthead:
Resolved by #856

PR #856, opened 2026-04-08
fix: Security parameter should come from genesis values

ChrisPaprocki:
## Description

This fixes using hardcoded k param. With the changes, the k param is fetched from genesis values and then from `cardano.protocol.parameters` as announced by `parameters_state`.

## Related Issue(s)
#818 

## How was this tested?
`make run` and look into logs for:

Logs:
```
2026-04-10T10:12:21.368755Z  INFO acropolis_module_consensus: Security parameter k set from genesis values k=2160
...
2026-04-10T10:12:21.368869Z  INFO acropolis_module_peer_network_interface::block_flow: Security parameter k set from genesis values security_param_k=2160
```

## Checklist

I'm convinced about handing k from `cardano.protocol.parameters`. I'm fairly sure k will not change there.

### DIFF
--- a/common/src/genesis_values.rs
+++ b/common/src/genesis_values.rs
@@ -20,6 +20,7 @@ pub struct GenesisValues {
     pub shelley_genesis_hash: Hash<32>,
     pub genesis_delegs: GenesisDelegates,
     pub magic_number: MagicNumber,
+    pub security_param: u64,
 }
 
 impl GenesisValues {
@@ -82,6 +83,7 @@ impl GenesisValues {
             ])
             .unwrap(),
             magic_number: MagicNumber::new(764824073),
+            security_param: 2160,
         }
     }
 
--- a/modules/consensus/src/consensus.rs
+++ b/modules/consensus/src/consensus.rs
@@ -932,41 +932,37 @@ impl ConsensusRuntime {
         force_validation: bool,
     ) {
         loop {
-            tokio::select! {
-                result = block_subscription.read() => {
-                    let Ok((_, message)) = result else {
-                        error!("Block message read failed");
-                        return;
+            let Ok((_, message)) = block_subscription.read().await else {
+                error!("Block message read failed");
+                return;
+            };
+
+            match message.as_ref() {
+                Message::Cardano((raw_blk_info, CardanoMessage::BlockAvailable(raw_block))) => {
+                    let block_info = if force_validation && self.do_validation {
+                        raw_blk_info.with_intent(BlockIntent::ValidateAndApply)
+                    } else {
+                        raw_blk_info.clone()
                     };
 
-                    match message.as_ref() {
-                        Message::Cardano((raw_blk_info, CardanoMessage::BlockAvailable(raw_block))) => {
-                            let block_info = if force_validation && self.do_validation {
-                                raw_blk_info.with_intent(BlockIntent::ValidateAndApply)
-                            } else {
-                                raw_blk_info.clone()
-                            };
-
-                            let span = info_span!("consensus", block = block_info.number);
-                            self.handle_block_available_direct(block_info, raw_block.clone())
-                                .instrument(span)
-                                .await;
-                        }
-
-                        Message::Cardano((
-                            _,
-                            CardanoMessage::StateTransition(StateTransitionMessage::Rollback(_)),
-                        )) => {
-                            self.context
-                                .message_bus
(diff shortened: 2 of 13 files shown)


common/src/genesis_values.rs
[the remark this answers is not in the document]
ChrisPaprocki: This is used only tests and should be moved to test utils. Lets do with #868

modules/consensus/src/consensus.rs
```diff
@@ -206,60 +214,121 @@ impl Consensus {
...
-            validator_topics,
-            validator_subscriptions,
-            validation_timeout,
-            do_validation,
-            stats: ConsensusStats::default(),
-        };
-
+        let run_context = context.clone();
         context.run(async move {
+            let genesis_k = Self::wait_genesis_values(&mut genesis_subscription)
```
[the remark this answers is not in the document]
ChrisPaprocki: Removed listening on protocol params

ChrisPaprocki:
> Thank you for the configuration info. After making those changes and executing `make run` I get this error:
> 
> ```
> 2026-04-09T19:31:06.222701Z  INFO acropolis_module_rest_blockfrost: Creating request handler on 'rest.get.txs.metadata.*'
> 2026-04-09T19:31:06.222712Z  INFO caryatid_process: Initialising module peer-network-interface
> Error: enum SyncPoint does not have variant constructor orgin for key `sync-point`
> ```
> 
> I don't recognize this error; do you have any suggestions on what I might have done wrong?

I could not reproduce this issue. After simplification, now only `make run` is needed.

buddhisthead:
Tested the branch and it works for me. I saw both expected logfile entries.

buddhisthead (approved):
Running the branch works for me. Nice work.

modules/consensus/src/consensus.rs
```diff
@@ -898,38 +932,41 @@ impl ConsensusRuntime {
...
-                return;
-            };
-
-            match message.as_ref() {
-                Message::Cardano((raw_blk_info, CardanoMessage::BlockAvailable(raw_block))) => {
-                    let block_info = if force_validation && self.do_validation {
-                        raw_blk_info.with_intent(BlockIntent::ValidateAndApply)
-                    } else {
-                        raw_blk_info.clone()
+            tokio::select! {
```
whankinsiv: It looks like this select can be removed

common/src/genesis_values.rs
```diff
@@ -82,6 +83,7 @@ impl GenesisValues {
             ])
             .unwrap(),
             magic_number: MagicNumber::new(764824073),
+            security_param: 2160,
```
shd, line 86: Let it stay, but I hate these constants in genesis values in the code.

shd (approved):
LGTM for now.

BUT: there was a discussion with Paul some time ago, that genesis parameters can be parsed by generic library. It's a bad idea to send data through pipes, if the data can be parsed globally inside one module (it's not good to send global constants from one module to another, parse genesis twice by different code in different modules, etc).

I see such change as temporary, and if there will be time/possibility to change, I'd made a separate common library ("genesis parameters"), and I'd restore these values from it. These parameters should be possible to configure in common section of omnibus.toml (incl. specifying separate parameters directly). 

(this is my special opinion on the issue)

common/src/genesis_values.rs
ChrisPaprocki: Turns out those are used in tests only, so likely they should be restructured as such, pls see #868.

ChrisPaprocki:
> LGTM for now.
> 
> BUT: there was a discussion with Paul some time ago, that genesis parameters can be parsed by generic library. It's a bad idea to send data through pipes, if the data can be parsed globally inside one module (it's not good to send global constants from one module to another, parse genesis twice by different code in different modules, etc).
> 
> I see such change as temporary, and if there will be time/possibility to change, I'd made a separate common library ("genesis parameters"), and I'd restore these values from it. These parameters should be possible to configure in common section of omnibus.toml (incl. specifying separate parameters directly).
> 
> (this is my special opinion on the issue)

If you refer to 'genesis_bootstrapper' and 'parameters_state' modules overlapping the parsing of genesis files, then I fully agree. 
Also if I read your comment correctly you also refer that genesis files shall not be parsed values flying around as messages, to which I also agree. As a part of another refactoring.
In this PR the downstream modules are consuming already parsed genesis files from 'genesis_bootstrapper'.

Issue #92, opened 2025-07-22
Add config helpers to common


sandtreader:
There a few different configuration reading helpers scattered about to fetch config values with a default - it would be good to centralise these in common/src/config.rs and then fix the config readers in all the modules to use them.

There is an emerging pattern of using a tuple of (config-property-name, default) in constants e.g. (in ParametersState)

const DEFAULT_STORE_HISTORY: (&str, bool) = ("store-history", false);

whankinsiv:
Closing with the merge of #863

PR #862, opened 2026-04-09
chore: cleanup accounts_state initialization 

whankinsiv:
## Description

This PR cleans up the accounts state initialization by:
* Moving configuration logic to a dedicated `configuration.rs` file
* Moving query handling logic to a dedicated `queries.rs` files

I've also added config helpers to common so that we will no longer need to define local helpers. 

## Related Issue(s)
Relates to #92 

## How was this tested?
* Verified that sync progresses as expected

## Checklist

## Impact / Side effects
N/A 

N/A


modules/accounts_state/src/state.rs
```diff
@@ -418,15 +418,9 @@ impl State {
...
-    /// Log statistics
-    fn log_stats(&self) {
-        info!(num_stake_addresses = self.stake_addresses.lock().unwrap().len());
-    }
-
     /// Background tick
-    pub async fn tick(&self) -> Result<()> {
-        self.log_stats();
-        Ok(())
+    pub fn tick(&self) {
```
ChrisPaprocki: Minor: `tick` was a bit confusing to me at first look. Pls consider renaming to something to log_stats (as was the helper before).

modules/accounts_state/src/accounts_state.rs
```diff
@@ -607,120 +588,23 @@ impl AccountsState {
...
-        if let Ok(verify_rewards_files) = config.get_string("verify-rewards-files") {
-            info!("Verifying rewards against '{verify_rewards_files}'");
-            verifier.set_rewards_template(&verify_rewards_files);
-        }
-
-        if let Ok(verify_spdd_files) = config.get_string("verify-spdd-files") {
-            info!("Verifying rewards against '{verify_spdd_files}'");
-            verifier.set_spdd_template(&verify_spdd_files);
-        }
+        let accounts_cfg = AccountsConfig::load(context.clone(), &config).await?;
```
ChrisPaprocki: I think this is doing more than getting config. Maybe indicate it is not only loading?

modules/accounts_state/src/state.rs
whankinsiv: Updated to `log_stats()` in 141fa87

modules/accounts_state/src/accounts_state.rs
whankinsiv: Changed to `init` in 141fa87

ChrisPaprocki (approved):

Issue #841, opened 2026-03-31
Validate ledger state modules rollback to correct point


whankinsiv:
The following modules will be validated by capturing their `State` at a particular block, advancing a number of blocks forward within K, and then rolling back to the block at which the state was captured. This will validate that our rollback logic is correct and works as intended. 

Modules:
- [x] utxo_state
- [x] spo_state
- [x] drep_state
- [x] governance_state
- [x] parameters_state
- [x] stake_delta_filter
- [x] epochs_state
- [x] accounts_state
- [x] chain_store
- [x] block_vrf_validator
- [x] block_kes_validator
- [x] tx_unpacker

buddhisthead:
@whankinsiv completed all but the governance_state, assigning to @shd for the last module.

PR #881, opened 2026-04-17
feat: add rollback support to governance_state

whankinsiv:
## Description

This PR refactors `governance_state` to use `StateHistory`, enabling rollback support. Additionally, this PR introduces `consume_sync_opt` on `ValidationContext`, simplifying the message read pattern in governance_state. This pattern should be used in all other state modules at some point in the future. 

## Related Issue(s)
Completes #841

## How was this tested?
* Verified that sync progresses as normal and rewards calculations match to tip on the boot from snapshot path. 

## Checklist

## Impact / Side effects
Increases memory usage of `governance_state` due to keeping the k block history in imbl structures

N/A


common/src/caryatid.rs
```diff
@@ -342,6 +342,17 @@ impl ValidationContext {
         inp
     }
 
+    pub fn consume_sync_opt<T>(
```
shd: Originally there were 'consume_sync' and simply 'consume'. Sync means something similar to 'primary' -- some additional synchronization functionality. So, from current perspective I'd remove 'sync' from both functions: just 'consume' and 'consume_opt'.

modules/governance_state/src/conway_voting.rs
```diff
@@ -693,15 +694,15 @@ impl ConwayVoting {
             self.reference_aggregates.get(&(new_epoch - 1, action_id.clone()))
         {
             if ref_v != &aggregated {
-                warn!("Verifying {action_id} at {new_epoch}: computed {aggregated} != reference {ref_v}");
+                error!("Verifying {action_id} at {new_epoch}: computed {aggregated} != reference {ref_v}");
```
shd, line 697: It's not an error yet.
What's the selection for the names in this module:
1. Error -- means that anyone who sees it must make an action. Error will be looked by others when looking for the reason for the deadlock
2. Warning -- something is wrong, but does not require immediate action.

This type of verification errors can happen regularily, and it'll overshadow real errors. I'd remain WARN here.

modules/governance_state/src/conway_voting.rs
```diff
@@ -693,15 +694,15 @@ impl ConwayVoting {
...
             } else {
-                info!("Verifying {action_id} at {new_epoch}: votes {aggregated}, ok");
+                debug!("Verifying {action_id} at {new_epoch}: votes {aggregated}, ok");
```
shd, line 699: This info is important, i'd leave it like that.

modules/governance_state/src/conway_voting.rs
```diff
@@ -693,15 +694,15 @@ impl ConwayVoting {
...
             }
 
             if ref_oc != &aggregated_outcome {
                 error!("Verifying voting outcome: {action_id} at {new_epoch}: computed {aggregated_outcome} != reference {ref_oc}");
             } else {
-                info!("Verifying voting outcome: {action_id} at {new_epoch}: {ref_oc}, ok");
+                debug!("Verifying voting outcome: {action_id} at {new_epoch}: {ref_oc}, ok");
```
shd, line 705: This is super-important INFO. 

Imagine I want to check that it's ok. I don't see any ERRORs. Does it say that votes are correct?
1. It can be just a configuration issue (no voting outcome files present: e.g. path is incorrect in .toml file)
2. It can be correct verification.

So, an event of verification (at least the final one) should be visible: either it ends with Error, or ends with success.

modules/governance_state/src/state.rs
```diff
@@ -274,4 +271,66 @@ impl State {
     pub fn get_conway_voting_mut(&mut self) -> &mut ConwayVoting {
         &mut self.conway_voting
     }
+
+    pub async fn process_drep_spo(
```
shd: It was a deliberate distibution: all external processing happens in governance-state.rs, all data storage happens in state.rs.

Why we should mix this in different files?

shd (changes requested):

common/src/caryatid.rs
whankinsiv: Updated to `consume` and `consume_opt` in 21cbcbc

modules/governance_state/src/conway_voting.rs
whankinsiv: I don't understand why this wouldn't be considered an error. If our computed votes are misaligned from the ledger reality then this means that we either have incorrect logic or received an invalid block which requires a rollback.

modules/governance_state/src/conway_voting.rs
whankinsiv: In my opinion we should limit log spam and remove most logging related to the happy path throughout the codebase. Ideally I think we would only have logging from the epoch state module which informs of a new epoch.

modules/governance_state/src/conway_voting.rs
whankinsiv: If we did not receive an error then its implied the votes are correct in my opinion.

modules/governance_state/src/state.rs
whankinsiv: I've moved `process_drep_spo` back to `governance_state.rs` in bece3a7.

modules/governance_state/src/conway_voting.rs
shd: It can be a false positive: configuration is wrong.
The most obvious situation is config param missing.

modules/governance_state/src/conway_voting.rs
shd: We have 'ERRORs' in treasury computation. They're told to be ignored: but definitely, something is not correctly computed inside. 

* There are known bugs, which can be tolerated at the moment
* And there is something to react

We deliver a code, which corresponds to some claim. If the claim is not fulfilled, ERROR signals about that. 

Any other logic is good, but it should be consistent over the project. We may discuss it today.

shd (approved):
LGTM

PR #892, opened 2026-04-23
chore: remove log spam from omnibus process

whankinsiv:
## Description

This PR replaces all info! logs with debug! in the omnibus process except `epoch_state`'s new epoch message. This makes it easier to inspect errors and follow the sync progress. I've also disabled governance verification by default as it creates significant log spam.

## Related Issue(s)
N/A

## How was this tested?
N/A

## Checklist

## Impact / Side effects
N/A

N/A

### DIFF
--- a/modules/accounts_state/src/verifier.rs
+++ b/modules/accounts_state/src/verifier.rs
@@ -18,7 +18,7 @@ use std::{
     fs::{read_dir, File},
     path::PathBuf,
 };
-use tracing::{debug, error, info, warn};
+use tracing::{debug, error, warn};
 
 /// Verifier
 pub struct Verifier {
@@ -273,7 +273,7 @@ impl Verifier {
             }
 
             if pots == desired_pots {
-                info!(epoch = epoch, "Verification success for");
+                debug!(epoch = epoch, "Verification success for");
             }
         } else {
             warn!("Epoch {epoch} not represented in verify test data");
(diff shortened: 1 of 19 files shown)


golddydev (approved):

modules/accounts_state/src/verifier.rs
```diff
@@ -520,7 +520,7 @@ impl Verifier {
 
         let (outcome, total, _, _, _) = Self::verify_spdd_impl(epoch, spdd, &reference);
         if outcome {
-            info!("Verification of SPDD, end of epoch {epoch}: OK, total active stake {total}");
+            debug!("Verification of SPDD, end of epoch {epoch}: OK, total active stake {total}");
```
shd: I think we should leave verification outcome. Let it be just one outcome for all verifications, but it's necessary.

E.g., I reference this info for acceptance critera: it's a sign that the process really doing something, not just runs through blocks.

shd (changes requested):
I like the whole idea, it's definitely worth it.

However, I still think that we need just 1-2 lines per epoch for internal verification outcomes.

modules/accounts_state/src/verifier.rs
whankinsiv: Moved SPDD verification back to `info!` in 402dadc.

shd (approved):
LGTM

PR #898, opened 2026-04-27
chore: Adjust logging levels for peers management.

ChrisPaprocki:
## Description

Peers management logs should log as `info` rather then `error` if the situation (e.g. disconnecting of discovered peer) is expected.

## Related Issue(s)
N/A

## How was this tested?
E.g. `run mainnet` or `run testnet` as specified in `MS-5-8-tests-v1` tag.

## Checklist

## Impact / Side effects
N/A

N/A

### DIFF
--- a/modules/peer_network_interface/src/network.rs
+++ b/modules/peer_network_interface/src/network.rs
@@ -568,7 +568,12 @@ impl NetworkManager {
             );
             return;
         };
-        warn!(address = %peer.conn.address, "disconnected from peer");
+
+        if self.configured_addrs.contains(&peer.conn.address) {
+            warn!(address = %peer.conn.address, "disconnected from pre-configured peer");
+        } else {
+            info!(address = %peer.conn.address, "peer disconnected");
+        }
 
         // Capture state before peer is partially consumed.
         let is_cold_origin = self.cold_origin.remove(&id);
@@ -592,7 +597,7 @@ impl NetworkManager {
         if is_cold_origin && !established && !is_configured {
             // Cold-promoted peer that never established a connection (TCP refused / timeout).
             // Blacklist so peer-sharing cannot re-add it this session.
-            warn!(
+            info!(
                 address = %address,
                 "cold-promoted peer never connected — blacklisting for session"
             );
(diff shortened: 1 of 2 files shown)


modules/peer_network_interface/src/network.rs
```diff
@@ -568,7 +568,12 @@ impl NetworkManager {
             );
             return;
         };
-        warn!(address = %peer.conn.address, "disconnected from peer");
+
+        if self.configured_addrs.contains(&peer.conn.address) {
+            warn!(address = %peer.conn.address, "disconnected from pre-configured peer");
+        } else {
+            info!(address = %peer.conn.address, "peer disconnected");
```
whankinsiv, line 575: Should this be debug?

modules/peer_network_interface/src/network.rs
```diff
@@ -592,7 +597,7 @@ impl NetworkManager {
         if is_cold_origin && !established && !is_configured {
             // Cold-promoted peer that never established a connection (TCP refused / timeout).
             // Blacklist so peer-sharing cannot re-add it this session.
-            warn!(
+            info!(
```
whankinsiv, line 600: Should this be debug?

modules/peer_network_interface/src/network.rs
```diff
@@ -568,7 +568,12 @@ impl NetworkManager {
             );
             return;
         };
-        warn!(address = %peer.conn.address, "disconnected from peer");
+
+        if self.configured_addrs.contains(&peer.conn.address) {
```
whankinsiv, line 572: Does it matter from a networking perspective if we disconnect from one of our preconfigured peers?

ChrisPaprocki: Yes. It is twofold.
On the operation level, the operator needs to know the pre-configurd peer is unstable (or that is just disconnected).
Then the pre-configured peers are treated differently - the node will try to reconnect to them. Though likely it should be with slight exponential backoff.

modules/peer_network_interface/src/network.rs
ChrisPaprocki: I'd keep it info. It is not warn, but also not debug. One might want to know how stable is the network and peers.

modules/peer_network_interface/src/network.rs
ChrisPaprocki: I'd keep info. This is metric for network health.

whankinsiv (approved):

